PR #4, opened 2023-08-02
DAS-1892 - Add pycodestyle check to automated test suite.

owenlittlejohns:
## Description

This PR stems from conversations with @eni-awowale during her work on DAS-1891. There wasn't an automated check for code styling in the regular bunch of unit tests. This PR adds that. (The test mimics others from repositories like HOSS).

## Jira Issue ID

DAS-1892 (not strictly, but sort-of related)

## Local Test Steps

N/A

## PR Acceptance Checklist
* ~~Documentation updated (if needed).~~

### DIFF
--- a/tests/test_code_format.py
+++ b/tests/test_code_format.py
@@ -1,0 +1,31 @@
+from pathlib import Path
+from unittest import TestCase
+
+from pycodestyle import StyleGuide
+
+
+class TestCodeFormat(TestCase):
+    """ This test class should ensure all earthdata-varinfo Python code adheres
+        to standard Python code styling.
+
+        Ignored errors and warnings:
+
+        * E501: Line length, which defaults to 80 characters. This is a
+                preferred feature of the code, but not always easily achieved.
+        * W503: Break before binary operator. Have to ignore one of W503 or
+                W504 to allow for breaking of some long lines. PEP8 suggests
+                breaking the line for a binary operator is more "Pythonic".
+
+    """
+    @classmethod
+    def setUpClass(cls):
+        cls.python_files = Path('varinfo').rglob('*.py')
+
+    def test_pycodestyle_adherence(self):
+        """ Ensure all code in the `varinfo` directory adheres to PEP8 defined
+            standards.
+
+        """
+        style_guide = StyleGuide(ignore=['E501', 'W503'])
+        results = style_guide.check_files(self.python_files)
+        self.assertEqual(results.total_errors, 0, 'Found code style issues.')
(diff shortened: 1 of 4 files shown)


varinfo/cf_config.py
```diff
@@ -125,8 +125,8 @@ class object.
         mission_matches = re.match(mission, self.mission) is not None
 
         short_name_matches = (
-            short_name is None or
-            re.match(short_name, self.short_name) is not None
+            short_name is None
+            or re.match(short_name, self.short_name) is not None
```
owenlittlejohns, line 129: The two changes in this file are styling things relating to the binary operator breaking stuff (W503 vs W504).

eni-awowale: Oh nice! So to test if our code adheres to style requirements we would do something `pycodestyle cmr_search.py` or can we just run `pycodestyle` in the root directory and it will check everything?

tests/test_code_format.py
```diff
@@ -0,0 +1,31 @@
...
+
+class TestCodeFormat(TestCase):
+    """ This test class should ensure all earthdata-varinfo Python code adheres
+        to standard Python code styling.
+
+        Ignored errors and warnings:
+
+        * E501: Line length, which defaults to 80 characters. This is a
+                preferred feature of the code, but not always easily achieved.
+        * W503: Break before binary operator. Have to ignore one of W503 or
```
eni-awowale, line 15: Not a nit or suggestion but what's this binary operator business about haha?

eni-awowale (approved):
I just had some questions but everything looks good to me!

tests/test_code_format.py
owenlittlejohns: [Here's the biblical reference](https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

An example if this is whether you should have:

```
if (
    my_value == 1.2345
    or my_other_value == 2.3456
):
    ...
```

Or:

```
if (
    my_value == 1.2345 or
    my_other_value == 2.3456
):
    ...
```

Ultimately the answer is that either are okay, so long as the code is consistently one or the other, with PEP008 stating a mild preference for the first option. `pycodestyle` is automatically configured to warn against _both_ options, which is unhelpful, so the test ignores the checking of the one that is considered slightly preferred.

varinfo/cf_config.py
owenlittlejohns: The style check is now folded into the main test suite, so if you run `make test` it will run `pycodestyle` over the whole `varinfo` directory as part of that command. That's how I found these stragglers. Having that test in place now means that we can't merge the PR until the code is all compliant according to `pycodestyle` because that will be a test failure.

If you wanted to run it separately, though, you could use `pycodestyle` in the command line, too. `pycodestyle varinfo/*py` or `pycodestyle varinfo/cmr_search.py` as you say.

tests/test_code_format.py
owenlittlejohns: Yeah - either is okay, but we should choose one and it throughout the project. Then I turned off the warning for the one we went with. So that second example will still trigger a warning, but the first one won't.

PR #6, opened 2023-08-09
DAS-1894

eni-awowale:
**Description**
Added functionality for downloading a granule locally from CMR

**Jira Issue ID**
DAS-1894

**Local Test Steps**

```
# Import environment types
from cmr import CMR_UAT

from varinfo.cmr_search import get_granules, get_granule_link, download_granule

granule_response = get_granules(concept_id='G1245662789-EEDTEST',
                                cmr_env=CMR_UAT,
                                token=token_uat) # Change token to your UAT token

link = get_granule_link(granule_response)
# Saves to your current directory and returns path of saved file
download_granule(link, token=token_uat)
# Uncomment and change out_directory to prefered path
# download_granule(link,  token=token_uat, out_directory='/Foo/User/selected/out/directory')
```

**PR Acceptance Checklist**
[ X] Jira ticket acceptance criteria met.
[ X] CHANGELOG.md updated to include high level summary of PR changes.
 VERSION updated if publishing a release.
[ X] Tests added/updated and passing.
 Documentation updated (if needed).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,13 @@ __pycache__/
 
 # Generated files
 stash
+downloaded_granules
+*.nc4
+*.nc
+*.grb
+*.he5
+*.h5
+*.hdf5
 
 # VS Code
 .vscode
--- a/.gitignore
+++ b/.gitignore
@@ -23,12 +23,12 @@ __pycache__/
 # Generated files
 stash
 downloaded_granules
-*.nc4
-*.nc
-*.grb
-*.he5
-*.h5
-*.hdf5
+/*.nc4
+/*.nc
+/*.grb
+/*.he5
+/*.h5
+/*.hdf5
 
 # VS Code
 .vscode
--- a/tests/unit/test_cmr_search.py
+++ b/tests/unit/test_cmr_search.py
@@ -1,20 +1,28 @@
-from unittest import TestCase, mock
-from unittest.mock import patch, MagicMock
+from shutil import rmtree
+from tempfile import mkdtemp
+from unittest import TestCase
+from unittest.mock import patch, Mock
 
-from cmr import (GranuleQuery, CMR_UAT, CMR_OPS)
+from cmr import (GranuleQuery, CMR_UAT)
 import requests
 from requests.exceptions import HTTPError
 
-
-from varinfo.cmr_search import (get_granules, get_granule_link, get_granule_name,
+from varinfo.cmr_search import (get_granules, get_granule_link,
                                 download_granule)
 from varinfo.exceptions import (CMRQueryException, MissingGranuleDownloadLinks,
-                                MissingPositionalArguments, RequestsException)
+                                MissingPositionalArguments, GranuleDownloadException)
 
 
 class TestQuery(TestCase):
     ''' A class for testing functions in cmr_search.
     '''
+    @classmethod
+    def setUp(self):
+        self.output_dir = mkdtemp()
+
+    def tearDown(self):
+        rmtree(self.output_dir)
+    
     @patch('varinfo.cmr_search.GranuleQuery', spec=GranuleQuery)
     def test_with_concept_id(self, granule_query_mock):
         ''' Test when `get_granules` is called with, `concept_id`,
@@ -37,7 +45,6 @@ class TestQuery(TestCase):
         query_response = get_granules(concept_id,
                                       cmr_env=CMR_UAT,
                                       token='foo')
-
         # Check if `get_granules` returns the CMR GranuleQuery output
         self.assertListEqual(query_response, granule_response)
 
@@ -250,48 +257,32 @@ class TestQuery(TestCase):
             self.assertEqual(get_granule_link(granule_response_links_correct),
                              'https://data.gesdisc.earthdata.nasa.gov/.nc4')
     
-    def test_granule_name(self):
-        ''' Check if get_granule_name returns expected string for response
-            with key "producer_granule_id" and raises IndexError for response
-            without a "producer_granule_id"
-        '''
-        granule_response = [{
-            'producer_granule_id': 'MERRA2_400.tavg1_2d_slv_Nx.20210601.nc4',
-            'boxes': ['-90 -180 90 180'],
-            'time_start': '2021-06-01T00:00:00.000Z'}]
-        
-        granule_response_no_producer_id = [{
-            'boxes': ['-90 -180 90 180'],
-            'time_start': '2021-06-01T00:00:00.000Z'}]
-        
-        with self.subTest('Granule has "producer_granule_id"'):
-            self.assertEqual(get_granule_name(granule_response),
-                             'MERRA2_400.tavg1_2d_slv_Nx.20210601.nc4')
-        
-        with self.subTest('Granule does not have "producer_granule_id"'):
-            with self.assertRaises(IndexError):
-                get_granule_name(granule_response_no_producer_id)
-    
     
     def test_download_gran_exceptions(self):
         ''' Check if exceptions are raised for incorrect input parameters
         '''
+    
+        with self.assertRaises(MissingPositionalArguments):
+            # Input parameter, token, is missing
+            download_granule(granule_link='https://foo.gov/example.nc4')
             
-        with self.assertRaises(ValueError):
-            # Input parameter, `granule_link` should be string
-            download_granule(5, 'foo')
-            
-        with self.assertRaises(ValueError):
-            # Input parameter, `out_filename` should be string
-            download_granule('foo', 10)
+        with self.assertRaises(NotADirectoryError):
+            # Input parameter, `out_directory` should be a real directory
+            download_granule('https://foo.gov/example.nc4', 
+                             '/Usr/foo/dir',
+                             'foo')
+        
+        with self.assertRaises(MissingPositionalArguments):
+            # Missing argument, `granule_link`
+            download_granule(token='foo')
         
         with self.assertRaises(ValueError):
-            # Input parameter, `out_filename` should be string
-            download_granule('foo', 10)
+            # `granule_link` should be str or bytes like not int
+            download_granule(granule_link=0, token='foo')
         
-        with self.assertRaises(RequestsException):
+        with self.assertRaises(GranuleDownloadException):
             # Not a real https link
-            download_granule('foo', 'foo.nc4')
+            download_granule('https://foo.gov/example.nc4', token='foo')
     
     
     def _mock_requests(self, status=200, content="CONTENT"):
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -22,6 +22,13 @@ __pycache__/
 
 # Generated files
 stash
+downloaded_granules
```
eni-awowale, line 25: `download_granules` saves a file into the current directory, so I added these file types to the `.gitignore`

tests/unit/test_cmr_search.py
```diff
@@ -1,11 +1,15 @@
-from unittest import TestCase
-from unittest.mock import patch
+from unittest import TestCase, mock
+from unittest.mock import patch, MagicMock
```
eni-awowale, line 2: Oops I don't use `MagicMock` will remove for the next commit

varinfo/cmr_search.py
```diff
@@ -62,13 +63,41 @@ def get_granules(concept_id: str = None,
     return granule_response
 
 
-def get_granule_link(query_response: list) -> str:
+def get_granule_link(granule_response: list) -> str:
```
eni-awowale, line 66: I thought this name was better

varinfo/cmr_search.py
```diff
@@ -62,13 +63,41 @@ def get_granules(concept_id: str = None,
...
+                         for link in granule_response[0].get('links', [])
                          if link['rel'].endswith('/data#') and 'inherited' not in link), None)
 
     if granule_link is None:
-        raise MissingGranuleDownloadLinks(query_response)
+        raise MissingGranuleDownloadLinks(granule_response)
     return granule_link
+
+
+def get_granule_name(granule_response: list) -> str:
```
eni-awowale, line 78: This really just a helper function for getting the granule filename

varinfo/cmr_search.py
```diff
@@ -62,13 +63,41 @@ def get_granules(concept_id: str = None,
...
+        # ie. MERRA2_100.tavg1_2d_slv_Nx.19800320.nc4
+    except Exception:
+        raise IndexError('No key named "producer_granule_id"')
+
+
+def download_granule(granule_link: str, out_filename: str):
+    ''' Use the requests module to download data via https.
+    '''
+    # Check if a string was entered for input parameters
+    if isinstance(granule_link, str) and isinstance(out_filename, str):
```
eni-awowale, line 92: I decided go with checking if `granule_link` and `out_filename` are strings. I won't be offended if you think we should do this differently.

tests/unit/test_cmr_search.py
```diff
@@ -238,4 +249,87 @@ def test_exceptions_granule_link(self):
...
+        with self.assertRaises(ValueError):
+            # Input parameter, `out_filename` should be string
+            download_granule('foo', 10)
+        
+        with self.assertRaises(RequestsException):
+            # Not a real https link
+            download_granule('foo', 'foo.nc4')
+    
+    
+    def _mock_requests(self, status=200, content="CONTENT"):
```
eni-awowale, line 297: Okay, I am pretty proud of this one. I wanted to mock `requests.get` and thought it would be easier to write a function that returns this mock object, instead of creating the mocked object in the same test for checking the mocked object against download_granules. Also, I noticed in some of your harmony code you start functions that are only used once with a `_` ?? I was reading somewhere that this is convention but feel free to correct me if I am wrong.

varinfo/cmr_search.py
```diff
@@ -62,13 +63,41 @@ def get_granules(concept_id: str = None,
...
+    ''' Syntax for extracting the granule name from the granule metadata.
+    '''
+    try:
+        return granule_response[0]['producer_granule_id']
+        # ie. MERRA2_100.tavg1_2d_slv_Nx.19800320.nc4
+    except Exception:
+        raise IndexError('No key named "producer_granule_id"')
+
+
+def download_granule(granule_link: str, out_filename: str):
```
eni-awowale, line 88: Oops I should probably also have type annotation for the return value....

owenlittlejohns (changes requested):
Nice work. In particular I really like the thorough test coverage.

Most of my stuff is minor, I think the main things are:

* Maybe swapping out the use of `producer_granule_id` for the `basename` of the URL. (We can chat about that if it doesn't sound right)
* Adding the `Authorization` header to your `requests.get` call.
* Possibly making your new function take in the directory to download to, and return the path of the local file you create.
* Update the test for the granule download so that you crack open the file you create and check it has the right contents.

william-valencia:
We should also update the version so I can pull the new changes down from pip.

owenlittlejohns:
@william-valencia - some quick thoughts on the need to release a new version of this package:

I'd been planning to hold off doing that until we had the full set of functionality that will be used by this API. Right now we don't have the function that lambda(s) within the API will call. That's will be created during DAS-1895, which isn't begun, yet.

The way the work is structured in the overall epic, the lambda(s) in cmr-graphql won't be looking to pull in any version of `earthdata-varinfo` until DAS-1901.

(Disclaimer - I wasn't in the meeting at the start of last week, so things might have changed. But this is how the tickets were intended when written)

varinfo/cmr_search.py
```diff
@@ -62,13 +63,41 @@ def get_granules(concept_id: str = None,
...
+        raise MissingGranuleDownloadLinks(granule_response)
     return granule_link
+
+
+def get_granule_name(granule_response: list) -> str:
+    ''' Syntax for extracting the granule name from the granule metadata.
+    '''
+    try:
+        return granule_response[0]['producer_granule_id']
+        # ie. MERRA2_100.tavg1_2d_slv_Nx.19800320.nc4
```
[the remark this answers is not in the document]
eni-awowale: Okay sounds good!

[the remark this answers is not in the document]
eni-awowale: I think I am just going to delete this function and add something in `download_granule` like `out_filename = os.path.basename(granule_link)`

tests/unit/test_cmr_search.py
```diff
@@ -238,4 +249,87 @@ def test_exceptions_granule_link(self):
         with self.subTest('Granule has `links` and all correct attributes'):
             self.assertEqual(get_granule_link(granule_response_links_correct),
                              'https://data.gesdisc.earthdata.nasa.gov/.nc4')
-    
\ No newline at end of file
+    
+    def test_granule_name(self):
```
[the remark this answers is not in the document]
eni-awowale: Thanks!! Learned from the best!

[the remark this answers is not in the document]
eni-awowale: I am gonna delete the tests for `granule_name` since I deleted it's function.

varinfo/exceptions.py
```diff
@@ -91,3 +91,12 @@ class MissingGranuleDownloadLinks(CustomError):
     def __init__(self, download_link):
         super().__init__('MissingGranuleDownloadLinks',
                          f'No links for granule record: {download_link}')
+
+
+class RequestsException(CustomError):
```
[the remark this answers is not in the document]
eni-awowale: Good point!

tests/unit/test_cmr_search.py
```diff
@@ -1,11 +1,15 @@
-from unittest import TestCase
-from unittest.mock import patch
+from unittest import TestCase, mock
```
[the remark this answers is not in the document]
eni-awowale: I don't think I am following. I did `from unittest.mock import patch` because I can't `import patch` without first importing mock

tests/unit/test_cmr_search.py
```diff
@@ -238,4 +249,87 @@ def test_exceptions_granule_link(self):
...
+        mock_response.status_code = status
+        mock_response.content = content
+        return mock_response
+
+    @patch('requests.get')
+    def test_download_granule(self, mock_requests_get):
+        ''' Check if `download_granules` returns the expected
+            status_code and content for the mocked response.
+        '''
+        link = 'https://harmony.uat.earthdata.nasa.gov'
```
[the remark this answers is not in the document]
eni-awowale: Yeah good point!

tests/unit/test_cmr_search.py
```diff
@@ -238,4 +249,87 @@ def test_exceptions_granule_link(self):
...
+            # Not a real https link
+            download_granule('foo', 'foo.nc4')
+    
+    
+    def _mock_requests(self, status=200, content="CONTENT"):
+        '''
+        Mock requests.get module and it's methods content and status.
+        Return the mock object.
+        '''
+        mock_response = mock.Mock()
```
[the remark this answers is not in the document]
eni-awowale: Sounds good, do you think you could elaborate on why I would do this? I just want to make sure I am following.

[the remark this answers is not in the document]
owenlittlejohns: I think this is mainly a distinction between the function being mocked and the object being returned. You are mocking `requests.get`, which makes a lot of sense - we don't want to attempt to make HTTP requests during the tests, and we want to be able to establish `requests.get` was called with the correct parameters.

Generally, it's nice to make the return value of the `Mock` as close to the real thing you'd expect to get as possible, in this case a `requests.Response` object. You can do that by creating a new instance of an object and setting its attributes to what your code expects, so that when you test the code it can process that object as though things are normal.

The tricky thing is that it turns out creating a `requests.Response` instance isn't as easy as it could be. I'm sure there's a way, but I've not found it from a quick bit of playing around. Given that it is a bit trickier, the short cut here is to use `mock_response = Mock(spec=requests.Response)`, so at least the output object from this function has all the right attributes and methods, even if it isn't really a `requests.Response`.

(Sorry for the essay!)

tests/unit/test_cmr_search.py
```diff
@@ -1,16 +1,28 @@
...
 
 class TestQuery(TestCase):
     ''' A class for testing functions in cmr_search.
     '''
+    @classmethod
+    def setUp(self):
+        self.output_dir = mkdtemp()
+
+    def tearDown(self):
+        rmtree(self.output_dir)
```
owenlittlejohns, line 24: I think this is fine, but I am often a little wary of trying to use `rmtree` without first checking the directory exists. If there's anything in the source code that might delete the working directory, you run the risk of `rmtree` failing because it can't find what you asked to be deleted. I don't _think_ that's an issue with these tests, but a way to guard against it would be to import `os.path.exists` and then have:

```
def tearDown(self):
    if exists(self.output_dir):
        rmtree(self.output_dir)
```

tests/unit/test_cmr_search.py
```diff
@@ -238,4 +256,78 @@ def test_exceptions_granule_link(self):
...
+        # Set the mock_response with the `_mock_requests` object content method
+        mock_response = self._mock_requests(content=mock_content)
+        # Set the return_value of `mock_requests_get` to mock_response
+        mock_requests_get.return_value = mock_response
+        file_path = download_granule(link, self.output_dir, token='foo')
+        # Check if `download_granule` was called once with expected parameters
+        mock_requests_get.assert_called_once_with(link, headers=
+                                                  {'Authorization':
+                                                      f'Bearer {"foo"}'},
+                                                  timeout=10)
```
owenlittlejohns, line 316: Nit: There's a little bit of wonky indentation here. Maybe:

```
        mock_requests_get.assert_called_once_with(
            link,
            headers={'Authorization': f'Bearer {"foo"}'},
            timeout=10
        )
```

Second nit: You don't really need the f-string in your authorisation header value, if you are just explicitly naming a string within the string. (I said string too many times in a row there...)

varinfo/cmr_search.py
```diff
@@ -75,29 +81,38 @@ def get_granule_link(granule_response: list) -> str:
...
+            (the default is the current directory).
+        * token: Earthdata Login (EDL) bearer_token
     '''
-    # Check if a string was entered for input parameters
-    if isinstance(granule_link, str) and isinstance(out_filename, str):
-        f = open(out_filename, 'w+')
+    if granule_link is not None:  # Check if input parameter was entered
+        if isinstance(granule_link, str):  # Check if granule link is a string
+            # Check if `out_directory` is a directory and create out_filename
+            if os.path.isdir(out_directory):
```
owenlittlejohns, line 96: First up - this is great thinking to catch the case of a non-existent directory.

What would you reckon to trying to create the directory if it doesn't already exist:

```
if not isdir(out_directory):
   <make directory>

out_file = os.path.join(out_directory, os.path.basename(granule_link))
```

Also, the nice thing about `os.path.join` above is that it will use the correct type of slash to concatenate file paths depending on the OS running the code - Windows file paths use `\` instead of `/`.

owenlittlejohns (commented):
I think this looks mostly good - I'll see what you think about my last couple of comments - feel free to push back if you think they are a bit unnecessary.

eni-awowale (commented):
I added all of your suggestions plus a couple Pylint things i noticed

varinfo/cmr_search.py
```diff
@@ -53,7 +63,7 @@ def get_granules(concept_id: str = None,
 
     except Exception as cmr_exception:
         # Use custom exception, CMRQueryException, if CMR fails
-        raise CMRQueryException(str(cmr_exception))
+        raise CMRQueryException(str(cmr_exception)) from cmr_exception
```
owenlittlejohns, line 66: Nice!

owenlittlejohns (approved):
For some reason I couldn't reply in a thread to the following comment:

> I needed to add this try/except so I could test if the input directories pointed to files. For example, ./foo.nc4 would raise a FileExistsError, since it is not a directory it passes the if not os.path.isdir(), but you cannot create a directory in a file. For my unittest, I have a fake directory ('/Usr/foo/dir/) and it would raise a FileNotFoundError since it is not a directory it passes the if not os.path.isdir() but that path does not exist so you can't create a file in it. I decided to create another CustomException (I know I am obssessed) to cover both of these test cases. Without this test my coverage dropped to 95 percent and we're all about that 💯. I hope you don't think this is overkill but I am this way now because of you (haha!)

That does sound like something I would say. I think this works. The other alternative would be to do something like:

```
if not os.path.exists(out_directory):
    os.mkdir(out_directory)
elif not os.path.isdir(out_directory):
    raise DirectoryCreationException(...)
```
To be honest, it pretty much just gets you the same thing, so that's probably just a difference in the functions the pair of us tend to use more often.

Thanks for making all the updates!

tests/unit/test_cmr_search.py
```diff
@@ -238,4 +256,78 @@ def test_exceptions_granule_link(self):
...
+    
+    def test_download_gran_exceptions(self):
+        ''' Check if exceptions are raised for incorrect input parameters
+        '''
+    
+        with self.assertRaises(MissingPositionalArguments):
+            # Input parameter, token, is missing
+            download_granule(granule_link='https://foo.gov/example.nc4')
+            
+        with self.assertRaises(NotADirectoryError):
```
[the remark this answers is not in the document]
owenlittlejohns: That does sound like something I would say.

PR #9, opened 2023-09-15
DAS-1895 - Add single overarching function to simplify end-user workflows

owenlittlejohns:
## Description

This PR combines the search, download and publish functionality under a single new function, such that an end-user (or API) has fewer commands to run in their workflow. A few other things have happened in this PR:

* The `token` parameter for `cmr_search.get_granules` and `cmr_search.download_granule` has been updated to `auth_header`, so we can just pass through an `Authorization` header that contains either an EDL bearer token, or a LaunchPad token.
* Tweaking to the generation of variable native IDs, to better ensure uniqueness and to avoid issues with slashes in hierarchical variable names.
* Some test tweaks (some are me and my OCD, some are probably more important)

## Jira Issue ID

DAS-1895

## Local Test Steps

* Clone this branch.
* Grab a bearer token from EDL in UAT.
* Pip install the dependencies for the package.

```
from varinfo import generate_collection_umm_var

auth_header = 'Bearer <your EDL bearer token>'

# This call will default to UAT, and won't try to publish
generated_umm_var = generate_collection_umm_var(<UAT collection concept ID>, auth_header)

# This should print a load of lovely JSON
print(generated_umm_var)

# This call will try to publish:
published_umm_var = generate_collection_umm_var(<UAT collection concept ID>, auth_header, publish=True)
```
Fun fact - that `auth_header` is still using an EDL bearer token. Currently that works for me (even though the future is LaunchPad tokens only for publication). I didn't have a collection without variables, so the mismatch between native IDs meant that I wasn't able to associate my published UMM-Var records to the collection, too, because there are already associated UMM-Var records with the same variable name, but a different native ID.

## PR Acceptance Checklist


VERSION
```diff
@@ -1 +1 @@
-1.0.1
+2.0.0
```
owenlittlejohns, line 1: I went with a major version bump due to the incompatible change from the `token` kwarg to `auth_header` in a couple of functions (even though those are lower level).

tests/unit/test_generate_umm_var.py
```diff
@@ -0,0 +1,276 @@
...
+        mock_granule_query.return_value.parameters.assert_called_once_with(
+            downloadable=True,
+            sort_key='-start_date',
+            concept_id=self.collection_concept_id
+        )
+
+        # Ensure the call to download the granule had correct parameters
+        mock_download_granule.assert_called_once_with(self.netcdf4_url,
+                                                      self.bearer_token_header,
+                                                      out_directory=ANY)
```
owenlittlejohns, line 87: These `ANY` statements are a bit lazy. The issue is that the temporary directory will change every test run. If I had a bit more time, I'd be tempted to write a custom matcher that ensures the string format is as expected. But that maybe is overkill. 🤷

tests/unit/test_generate_umm_var.py
```diff
@@ -0,0 +1,276 @@
...
+
+        # Ensure the output looks as expected
+        # First check all the expected concept IDs are present:
+        self.assertEqual(set(concept_ids),
+                         set(published_umm_var).intersection(concept_ids))
+
+        # Next check the error matches the expected format,
+        # e.g.: '/variable_name: Error Message'.
+        # This is done in pieces, in case dictionary ordering means the
+        # variable associated with the error varies between test runs.
```
owenlittlejohns, line 250: This is all a bit ugly. It's primarily to guard against the chance that internally the order of variables in variable `VarInfoFromNetCDF4` related dictionaries varies. (Might be no longer an issue with Python ≥ 3.9, but still a bit wary from previous experience with it)

varinfo/cmr_search.py
```diff
@@ -15,44 +15,57 @@
                                 DirectoryCreationException)
 
 
+CmrEnvType = Literal[CMR_OPS, CMR_UAT, CMR_SIT]
```
owenlittlejohns, line 18: I vaguely remember a conversation from a prior PR about wanting a type that shows the CMR environments aren't just any strings, but are specific strings. I did a bit of digging around, and that's what `typing.Literal` seems to do. (Was pretty happy with this find!)

varinfo/cmr_search.py
```diff
@@ -15,44 +15,57 @@
...
+                 collection_shortname: str = None,
                  collection_version: str = None,
                  provider: str = None,
-                 cmr_env: CMR_OPS = CMR_OPS,
-                 token: str = None) -> list:
-    ''' Search CMR given:
-        * concept_id: a collection or granule's IDs
-        * shortname/short_name: 'SNDRSNIML2CCPRET' or 'M2T1NXSLV'
+                 cmr_env: CmrEnvType = CMR_OPS,
+                 auth_header: str = None) -> list:
```
owenlittlejohns, line 26: This is really the important change in this function signature (`token` to `auth_header`). It makes this function agnostic to EDL bearer tokens vs LaunchPad, vs whatever else comes in the future.

varinfo/generate_umm_var.py
```diff
@@ -0,0 +1,92 @@
+''' This module generates UMM-Var records for a specified collection, and
```
owenlittlejohns, line 1: This is the main bit of work that I've done. It's nothing too fancy, just wrapping existing functions and bringing them together in one place.

varinfo/generate_umm_var.py
```diff
@@ -0,0 +1,92 @@
...
+          been generated for.
+        * cmr_env: URLs for CMR environments (OPS, UAT, and SIT)
+        * auth_header: Authorization HTTP header, containing either:
+          - A LaunchPad token: 'Authorization: <token>'
+          - An EDL bearer token: 'Authorization: Bearer <token>'
+        * publish: Optional argument determining whether to publish the
+          generated UMM-Var records to the indicated CMR instance. Defaults to
+          False.
+
+        Note - if attempting to publish to CMR, a LaunchPad token must be used.
```
owenlittlejohns, line 46: As discussed in the PR description - it turns out this isn't currently true. But it likely will be soon, so I didn't want to confuse matters for future users.

varinfo/generate_umm_var.py
```diff
@@ -0,0 +1,92 @@
...
+
+    return return_value
+
+
+def is_variable_concept_id(possible_concept_id: str) -> bool:
+    """ A helper function to identify if a given string conforms to the
+        expected structure of a variable concept ID.
+
+    """
+    return bool(re.match(r'^V\d{10}-\w+$', possible_concept_id))
```
owenlittlejohns, line 92: I think tying this to 10 digits is okay. But at some point CMR will require 11. But that seems like a future issue maybe. Thoughts?

varinfo/umm_var.py
```diff
@@ -346,53 +346,74 @@ def get_json_serializable_value(input_value: Any) -> Any:
...
         'Authorization': auth_header,
-        'Accept': 'application/json'}
+        'Accept': 'application/json'
+    }
+    variable_native_id = generate_variable_native_id(collection_id,
+                                                     umm_var_dict)
 
     url_endpoint = (cmr_env.replace('search', 'ingest') + 'collections/'
-                    f'{collection_id}/variables/{umm_var_dict["Name"]}')
+                    f'{collection_id}/variables/{variable_native_id}')
```
owenlittlejohns, line 371: When I was testing the new function locally, I was getting a n error on publication where CMR couldn't find the endpoint we were looking for. Turns out it's because my test collection was hierarchical, so:

```
https://cmr.uat.earthdata.nasa.gov/ingest/collections/<collection concept ID>/variables/<native ID>
```

Looks something like:

```
https://cmr.uat.earthdata.nasa.gov/ingest/collections/C1234567890-PROV/variables/Grid/time
```

That extra slash was confusing CMR, making it thing we were looking for a different endpoint.

varinfo/umm_var.py
```diff
@@ -346,53 +346,74 @@ def get_json_serializable_value(input_value: Any) -> Any:
...
         return response.json()['concept-id']
-    # A failed request returns the response dict with the error message
-    # e.g. {'errors': ['#: required key [LongName] not found']}
-    return response.json()
+    else:
+        # A failed request returns the response dict with the error message
+        # e.g., {'errors': ['#: required key [LongName] not found']}
+        # This will be converted into a single string that can be returned to
+        # the end-user.
+        return '\n  '.join(response.json()['errors'])
```
owenlittlejohns, line 387: Aside from my OCD moving the comments around - I updated this to only return the error messages themselves (concatenated together into a single string).

varinfo/umm_var.py
owenlittlejohns: Using the collection concept ID as part of the native ID is a bit less human-friendly, but it's the only collection information was available to the function to ensure uniqueness within the provider. Also, I think it's generally only metadata curators that tend to see/use the native ID.

eni-awowale:
I am sorry I have to nit you on this @owenlittlejohns but you missed an excellent opportunity to title your PR "One Function to Rule them All!"

.github/workflows/run_tests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
 
     steps:
-      - name: Checkout harmony-gdal-adapter repository
+      - name: Checkout earthdata-varinfo repository
```
eni-awowale, line 15: Whoaa what does this do?

tests/unit/test_cmr_search.py
```diff
@@ -18,193 +18,230 @@
 
 
 class TestQuery(TestCase):
-    ''' A class for testing functions in cmr_search.
-    '''
+    ''' A class for testing functions in cmr_search. '''
     @classmethod
+    def setUpClass(cls):
```
eni-awowale, line 23: Really like this! It get's a little annoying retyping this for all the tests

.github/workflows/run_tests.yml
owenlittlejohns: This change just fixes where I copied over the CI/CD from the HGA repo and missed changing a reference from HGA to earthdata-varinfo. In particular, you can see the name of the step of this workflow change (see the second step in the runs [before](https://github.com/nasa/earthdata-varinfo/actions/runs/6194333475/job/16817159230) vs [after](https://github.com/nasa/earthdata-varinfo/actions/runs/6194450426/job/16817463275)).

Not sure if you've played around much with GitHub Actions before. If not, you can get your CI/CD to run in GitHub by defining it in YAML files that sit in `.github/workflows`. (I _really_ like having it all built into GitHub, and having the CI/CD under version control. Plus, they are just fun to play around with, even if debugging can be a bit of a pain)

owenlittlejohns: (I guess to clarify, this isn't a functional change, just essentially changing a label for a step in the workflow)

tests/unit/test_cmr_search.py
owenlittlejohns: :100:

I love these methods on `TestCase`. There are times when they save _soooo_ many lines of code.

varinfo/generate_umm_var.py
```diff
@@ -0,0 +1,92 @@
...
+        all_umm_var_records = get_all_umm_var(var_info)
+
+    if publish:
+        # Publish to CMR and construct an output object that is a list of
+        # strings. These strings will be either variable concept IDs or
+        # error messages returned from CMR.
+        publication_response = publish_all_umm_var(collection_concept_id,
+                                                   all_umm_var_records,
+                                                   auth_header, cmr_env)
+
```
eni-awowale, line 74: So for a collection if we have one variable fail ingesting and one pass this is what it looks like: 
`[
  {'/TROPPV': "['#: required key [LongName] not found']", 
  {'/TROPQ': 'V1259791548-EEDTEST'}
  ]`
And can you add whatever it is in the comments like `# e.g. ...`

varinfo/generate_umm_var.py
```diff
@@ -0,0 +1,92 @@
...
+        return_value = [variable_response
+                        if is_variable_concept_id(variable_response)
+                        else ': '.join([variable_name, variable_response])
+                        for variable_name, variable_response
+                        in publication_response.items()]
+    else:
```
eni-awowale, line 80: Why not just return `all_umm_var_records` I don't see what the point of making it into a list is.

README.md
```diff
@@ -98,6 +98,45 @@ umm_var = get_all_umm_var(var_info)
 export_all_umm_var_to_json(list(umm_var.values()), output_dir='local_dir')
 ```
 
+### End-to-end UMM-Var generation and publication:
+
+```
+from cmr import CMR_OPS
+from varinfo import generate_collection_umm_var
```
flamingbear, line 105: Is this actually importable from the package.  I didn't see it in __init__.py and I had to import it from varinfo.generate_umm_var

When I built the package and installed with `pip install -e .`. I got this error

```
>>> from varinfo import generate_collection_umm_var
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'generate_collection_umm_var' from 'varinfo' (/Users/savoie/projects/data-services/earthdata-varinfo/varinfo/__init__.py)
```

varinfo/generate_umm_var.py
eni-awowale: Might be a little overkill but we could also do a CMR `VariableQuery` search instead. if that returns true we know it's for sure a variable but I guess this is fine for now.

eni-awowale: Also do you mind defining `is_variable_concept_id` before using it?

tests/unit/test_cmr_search.py
```diff
@@ -18,193 +18,230 @@
...
+        # environment, Authorization header, and expected page size
         granule_query_mock.assert_called_once_with(mode=CMR_UAT)
-        granule_query_mock.return_value.bearer_token.assert_called_once_with
-        ('foo')
+        self.assertDictEqual(granule_query_mock.return_value.headers,
+                             {'Authorization': self.bearer_token_header})
         granule_query_mock.return_value.get.assert_called_once_with(10)
 
+    @patch('varinfo.cmr_search.GranuleQuery', spec=GranuleQuery)
+    def test_with_launcpad_token(self, granule_query_mock):
```
flamingbear, line 112:
```suggestion
    def test_with_launchpad_token(self, granule_query_mock):
```

tests/unit/data/f16_ssmis_20210426v7.nc
eni-awowale, line 1: Is this name of granule with its native-id? If not would you mind adding the granule native-id?

tests/unit/test_cmr_search.py
```diff
@@ -18,193 +18,230 @@
...
+
+        # Check if `get_granules` returns the CMR GranuleQuery output
+        self.assertListEqual(query_response, granule_response)
+
+        # Check if the `GranuleQuery` object was instantiated with the correct
+        # environment, Authorization header, and expected page size
+        granule_query_mock.assert_called_once_with(mode=CMR_UAT)
+        self.assertDictEqual(granule_query_mock.return_value.headers,
+                             {'Authorization': self.launchpad_token_header})
+        granule_query_mock.return_value.get.assert_called_once_with(10)
```
flamingbear, line 141: Is this doing anything different from the test up on L68-70? Or just explicitly testing that 'Bearer' doesn't need to be involved?

tests/unit/test_cmr_search.py
```diff
@@ -301,21 +340,52 @@ def test_download_granule(self, mock_requests_get):
...
+            link,
+            headers={'Authorization': self.bearer_token_header},
+            timeout=10
+        )
+        # Check if download file contains expected content from `requests.get`
+        with self.subTest('Test if downloaded file contains expected content'):
+            with open(file_path, 'r', encoding='utf-8') as file:
+                data = file.read()
+
+            self.assertEqual(data, link)
```
flamingbear, line 356: Is the content of these downloads usually a link? if not then this is a little confusing.

tests/unit/test_generate_umm_var.py
```diff
@@ -0,0 +1,276 @@
...
+        self.assertSetEqual(set(published_umm_var),
+                            set(expected_concept_ids))
+
+    @patch('varinfo.umm_var.publish_umm_var')
+    @patch('varinfo.cmr_search.GranuleQuery')
+    @patch('varinfo.generate_umm_var.download_granule')
+    def test_error_from_search_is_raised(self, mock_download_granule,
+                                         mock_granule_query,
+                                         mock_publish_umm_var):
+        """ Ensure an error raised iduring search is propagated out to the
```
eni-awowale, line 160: Typo with `during` and some other stuff... I think you wanted to say:
`""" Ensure an error is raised during search and it is propagated out to the ...`

flamingbear, line 160:
```suggestion
        """ Ensure an error raised during search is propagated out to the
```

tests/unit/test_generate_umm_var.py
```diff
@@ -0,0 +1,276 @@
...
+            response with an error.
+
+        """
+        error_message = 'Invalid JSON'
+        concept_ids = [
+            'V0000000001-PROV', 'V0000000002-PROV', 'V0000000003-PROV',
+            'V0000000004-PROV', 'V0000000005-PROV', 'V0000000006-PROV',
+            'V0000000007-PROV'
+        ]
+        concept_ids_and_error = concept_ids + [error_message]
```
eni-awowale, line 204: I am having a hard time visualizing what this will look like can you either add a `# e.g. ...` or fully declare `concept_ids_and_error` with `error_message` embedded?

tests/unit/test_generate_umm_var.py
eni-awowale: Yeah tbh this is a little too obfuscated for me...

flamingbear: I followed to here, I think there's only one error message but I don't know what it looks like going in to the stuff below. Meaning I don't see how there's a `/variable_name` on it but I guess I trust you

eni-awowale (changes requested):
Looks great! I left some detailed comments for the aspects I need more clarification on. 
Namely, Bearer token vs Launchpad Token (is a Bearer token still a bearer token without `Bearer` in front or does it become a Launchpad token?) and the variable native-ids with concept-ids in their names stuff.

tests/unit/test_generate_umm_var.py
flamingbear: Ok, I'm guessing this is what I hadn't seen yet https://github.com/nasa/earthdata-varinfo/pull/9/files#diff-58979b97c948a78fe40f778dac62d374aba90d16408ddfd23a8829338464bf22R75-R79

flamingbear: And I see that @eni-awowale has issues understanding also, maybe concrete examples in the comment?

varinfo/generate_umm_var.py
flamingbear: 😬 will they require 11 or will it be 10 or 11 and how long in the future, is this future owen problem or some new sap?

flamingbear (commented):
This is nice lots of good stuff.  I found a few typos that should be changed.  One test that was hard for me to understand. And just a couple of other comments and questions.  I'm taking off so I'll just comment.

varinfo/generate_umm_var.py
owenlittlejohns: I think we agreed with William and Ryan that they are expecting a bunch of concept IDs, and that we could add a verbose mode later on that would include both.

This list comprehension is doing a little bit more, though. If the message gotten back from `publish_umm_var` for the variable is indeed a variable concept ID, then that's what is returned as the list element. If there was an error, then the error message is combined with the variable long name to give the user a clearer indication of what failed. So, it'd look something like:

```
return_value = [
    'V1234567890-EEDTEST',
    '/full/variable/path: This is the error message from CMR'
]
```

varinfo/umm_var.py
```diff
@@ -346,53 +346,74 @@ def get_json_serializable_value(input_value: Any) -> Any:
...
+def generate_variable_native_id(collection_concept_id: str,
+                                umm_var_record: Dict) -> str:
+    """ A helper function to create a CMR native ID given the collection
+        concept ID and the variable UMM-Var JSON. This native ID must be unique
+        within the entire provider. The initial implementation will be to
+        concatenate the collection concept ID and the long name of the variable
+        while removing slashes that CMR will interpret as part of the URL path.
+
+    """
+    return '-'.join([collection_concept_id,
```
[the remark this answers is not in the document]
owenlittlejohns: I checked in with Ryan, as knower of all things CMR. He is wary of using the collection entry ID, short name and version in the variable (or even collection) native ID, as they are mutable fields. Recommendations were:

* use source (varinfo) and something like base64 of the concept-id
* or, just use varinfo and concept id. It's all programatic, so ensure whatever you do, code can determine what it is and how to find it

I think I'm feeling better about sticking with concept ID now.

varinfo/generate_umm_var.py
owenlittlejohns: (Mentioned in other thread - will add clarifying comment to make this easier to grok in the code)

varinfo/generate_umm_var.py
owenlittlejohns: Close, but the output for a response with an error is a bit different. It would look like:

```
[ {'/TROPPV': '#: required key [LongName] not found', {'/TROPQ': 'V1259791548-EEDTEST'} ]
```

If `/TROPPV` returned multiple errors, those get combined.

E.g.: CMR response JSON:

```
{
  "errors": ["error 1", "error 2"]
  ...
}
```

Combined output from `publish_all_umm_var`:

```
[{'/TROPPV': 'error 1\n  error 2', {'/TROPQ': 'V1259791548-EEDTEST'} ]
```

- [ ] I'll review the comments and documentation string for `publish_umm_var` to try better describe that output.

tests/unit/test_generate_umm_var.py
owenlittlejohns: - [ ] Fix typo

tests/unit/data/f16_ssmis_20210426v7.nc
owenlittlejohns: This is the name of the GHRC granule as saved by the DAAC.

I'm not clear on the request here - are you asking me to rename the file to use the native ID? I think it's more appropriate to keep it as the DAAC-chosen name for the file.

varinfo/generate_umm_var.py
owenlittlejohns: I think we're at something like C2XXXXXXXXX-PROV now, so I think we're talking _years_. Strongly tempted to leave as is and close the can of worms I just opened. (Deal with it Future Owen)

tests/unit/test_generate_umm_var.py
owenlittlejohns: As in beyond:

```
# e.g.: '/variable_name: Error Message'.
```

Or do you mean the comments in the actual code, rather than the tests? (If comments in the code, for sure, will try to improve)

README.md
owenlittlejohns: Yeah - good catch. I had added it temporarily to `__init__.py` so this would work. But then it turns out that caused a circular import issue and I reverted it (but not the documentation).

- [ ] Fix README

tests/unit/test_generate_umm_var.py
owenlittlejohns: - [ ] (I can make that test more commented, too, though)

varinfo/cmr_search.py
```diff
@@ -15,44 +15,57 @@
...
+        * shortname/short_name: e.g., 'SNDRSNIML2CCPRET' or 'M2T1NXSLV'
         * collection_version/version: a collection version: '2' or 5.12.4'
         * provider: data center ID (e.g. GES_DISC, PODAAC, POCLOUD, EEDTEST)
         * cmr_env/mode: CMR environments (OPS, UAT, and SIT)
-        * token: Earthdata Login (EDL) bearer_token
-    For a successful search response concept_id
-    or short_name, version and provider must be entered
-    along with a bearer_token.
+        * auth_header: Authorization HTTP header, containing either:
+          - A LaunchPad token: 'Authorization: <token>'
```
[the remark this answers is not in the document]
owenlittlejohns: I'm not sure I see the other comment.

tests/unit/test_cmr_search.py
owenlittlejohns: Yeah - this is an almost identical test, just a different type of token (without 'Bearer'). Prior to this change there was some bearer token specific stuff that would try to take the input and prepend 'Bearer ' to it.

owenlittlejohns: Possibly this could be parameterised... will look into.

tests/unit/test_cmr_search.py
```diff
@@ -18,193 +18,230 @@
...
+        # Check if the `GranuleQuery` object was instantiated with the correct
+        # environment, Authorization header, and expected page size
         granule_query_mock.assert_called_once_with(mode=CMR_UAT)
-        granule_query_mock.return_value.bearer_token.assert_called_once_with
-        ('foo')
+        self.assertDictEqual(granule_query_mock.return_value.headers,
+                             {'Authorization': self.bearer_token_header})
         granule_query_mock.return_value.get.assert_called_once_with(10)
 
+    @patch('varinfo.cmr_search.GranuleQuery', spec=GranuleQuery)
```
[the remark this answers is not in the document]
owenlittlejohns: Ah - for some reason I missed this. I see it now.

I think you've caught a good point here. I've slightly mixed up my terminology between "token" and "header".

- [ ] Clean up token vs header.

tests/unit/test_cmr_search.py
owenlittlejohns: - [ ] Improve download tests.

tests/unit/data/f16_ssmis_20210426v7.nc
eni-awowale: So in a UMM-G record the native-id is the same as the filename that gets saved when you call `cmr_search.download_granule`, so `native-id == out_filename`. I was unfamiliar with this native-id so i was confused, so no worries!

owenlittlejohns: [I'm not sure that convention is adopted across all DAACs](https://cmr.earthdata.nasa.gov/search/granules.umm_json?concept_id=G2560583706-NSIDC_ECS) (although sounds reasonable).

varinfo/umm_var.py
```diff
@@ -346,53 +346,74 @@ def get_json_serializable_value(input_value: Any) -> Any:
...
     For a successful requests all of these fields must be entered
     """
     return {
         var_name: publish_umm_var(collection_id, umm_var, auth_header, cmr_env)
-        for var_name, umm_var in all_umm_var_dict.items()}
+        for var_name, umm_var in all_umm_var_dict.items()
+    }
+
+
+def generate_variable_native_id(collection_concept_id: str,
```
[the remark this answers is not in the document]
eni-awowale: Oh whoops I think you misunderstood me, I just meant can you switch the order of 'generate_variable_native_id` and `publish_umm_var` so it comes first. Since you are using 'generate_variable_id` in `publish_umm_var`. By "defining" i just meant where you literally define the function not adding a doc string. Sorry that was unclear.
So this:
```
def generate_variable_native_id(....)
....
def publish_umm_var(...)
```

tests/unit/test_cmr_search.py
owenlittlejohns: ...was thinking more of some of the bells and whistles that you get with `pytest`. Going to give in to laziness here.

varinfo/umm_var.py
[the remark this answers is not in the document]
owenlittlejohns: Ohhhh. I understand now. I think I'm going to push back on this one. I can do that, sure.

varinfo/generate_umm_var.py
owenlittlejohns: > Also do you mind defining is_variable_concept_id before using it?

I understand this better now from your other comment.

I think with this module it makes the most sense to have `generate_collection_umm_var` at the top, because that's the main function in the module that we want an end user to interact with, so it should probably be front and centre (ha - neither of those words are synonyms for "at the top", but you know what I mean).

varinfo/umm_var.py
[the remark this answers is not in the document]
eni-awowale: > A native-id is an identifier, unique per provider, used to identify concepts within CMR. The native-id is a string with no specific pattern.

It feels like for consistency sake their _should_ be a specific pattern, but if it's allowed....

varinfo/umm_var.py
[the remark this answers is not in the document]
eni-awowale: Not sure if you are or not... In any case that was what I was talking about in `generate_umm_var.is_variable_concept_id` when I said:
 > Also do you mind defining is_variable_concept_id before using it?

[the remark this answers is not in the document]
owenlittlejohns: Ha - yeah sorry for the confusion! I moved `generate_variable_native_id` up and I think I pushed that commit. I made a comment on the `is_variable_concept_id` and we can track that conversation separately, if that sounds good.

varinfo/umm_var.py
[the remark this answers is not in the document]
owenlittlejohns: Yeah. Definitely some ickiness with there being no formal convention.

Chatting with Ryan about this was interesting. From his recommendations above, it feels like the convention ought to involve concept ID in some way, where that's immutable.

eni-awowale (approved):
Thanks for adding more comments and clarifying things for me!

PR #10, opened 2023-09-20
Das 1976

eni-awowale:
## Description

Added the notebook, `how_to_publish_to_cmr_with_earthdata_varinfo.ipynb`,
to the documentation directory

## DAS-1976
I updated to include all of the suggestions mentioned in the review

## PR Acceptance Checklist
* ~[ ] Jira ticket acceptance criteria met.~
*  The rest are N/A:
* [ ] `CHANGELOG.md` updated to include high level summary of PR changes.
* [ ] `VERSION` updated if publishing a release.
* [ ] Tests added/updated and passing.
* [ ] Documentation updated (if needed).


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## v2.0.1
```
owenlittlejohns, line 1: I think we should probably not include this in the release notes, as the documentation doesn't get bundled up and released to PyPI (it's only the stuff in the `varinfo` directory that gets included in a release).

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+   "source": [
+    "### Overview:"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This version of `earthdata-varinfo` adds functionality to publish records to CMR, along with a single overarching function that wraps the search, download and publication functionality into a single function for the convenience of the end-user. Additionally, the function signatures for `cmr_search.get_granules` and `cmr_search.download_granule` have been updated to accept the full Authorization header, instead of a bearer token, so that they are also compatible with LaunchPad tokens."
```
owenlittlejohns, line 32: This text look familiar!

I think it's probably worth reframing this introduction to the notebook more for telling an end-user what they will learn by working through it, rather than the release notes for v2.0.0. (Although, it's definitely a great call to mention that the version of `earthdata-varinfo` being used needs to be at least 2.0.0 to run through the notebook)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Setting up your environment to run this notebook"
```
owenlittlejohns, line 40: It's probably worth a quick sentence to mention that a user should initially set up and activate a Python environment with either `pyenv` or conda before running any `pip` commands.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "                auth_header=auth_header,\n",
+    "                cmr_env=CMR_UAT)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Download a granule for the collection **GLDAS_NOAH10_3H**, generate it's UMM-Var entry and publish it to CMR, with our one function: ```generate_collection_umm_var```"
```
owenlittlejohns, line 277: The flow of this notebook makes a lot of sense from the perspective of the order we did development. But this function is kind of the all-singing-all-dancing thing that we're trying to offer to folks to make their lives as easy as possible. Because of that, I reckon we should have this before all the individual function calls. Sort of the TL;DR workflow, you know?

Nitpick: This is a pretty long heading. I think that probably suggests that the heading could be shorter and the details expanding into some regular-text sentences in the markdown block.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "# publishing a UMM-Var entry\n",
+    "from varinfo.umm_var import (get_all_umm_var, get_umm_var, publish_all_umm_var,\n",
+    "                             publish_umm_var)\n",
+    "# Singular function call\n",
+    "from varinfo.generate_umm_var import generate_collection_umm_var"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
```
owenlittlejohns, line 124: Nit - the top two code cells have been committed in an executed state.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Add publish=True if you want publish directly to CMR UAT\n",
+    "generate_collection_umm_var(collection_concept_id='C1256543837-EEDTEST',\n",
+    "                            auth_header=auth_header)"
```
owenlittlejohns, line 288: It'd be good to include some truncated expected output in the markdown cell above this call - just so the user knows what to expect, and if the code ran correctly. (I think that's applicable to a lot of the code cells in the notebook)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "from varinfo.generate_umm_var import generate_collection_umm_var"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Update authorization header to include your bearer token (e.g. 'Bearer TOKEN')\n",
```
owenlittlejohns, line 128: Tokens and `Authorization` headers are probably the trickiest part of this whole notebook. I think it would be really helpful to have a markdown cell describing this in a bit of detail.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "### Then download the granule locally with `download_granule`"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Saves granule to current directory or\n",
```
owenlittlejohns, line 168: This is a general comment that applies to a lot of the code-cells...

The comments in the code-cells are really informative, but there's quite a few places where I think the comments from the code could/should be migrated into markdown (and probably expanded a bit). That's particularly true for code-cells that are executing only one line of code.

My general thought here is that someone looking at a documentation notebook is probably going to start by reading the markdown sections first, rather than comments within code.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "# out_directory=/usr/given/path\n",
```
owenlittlejohns, line 169: Nit: A lot machines actually have a `/usr` directory, so this example directory is a tiny bit misleading.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "umm_var_dict"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Publish all the UMM-Var entries for M2I1NXASM"
+   ]
```
owenlittlejohns, line 218: What should the user expect to see when this code is completed?

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "\n",
+    "`earthdata-varinfo` is also able to create UMM-Var compliant JSON records and publish these records to NASA's Common Metadata Repository (CMR)."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Example usage:\n",
```
owenlittlejohns, line 92: These examples work for folks with the correct privileges for the EEDTEST CMR provider in UAT. I think there's a little bit of missing text prompting the user to set the collection concept ID to a collection they have the correct privileges for.

Maybe there should be a cell where the collection concept ID is set early on (kind of like the `Authorization` header is now) and then that same variable is called in all the later cells? Also, I think we only need one collection here, for simplicity.

owenlittlejohns (changes requested):
I think the scope of all the code blocks is exactly what we need to be explaining in this notebook - nice!

My comments are mainly about trying to make this notebook a little bit more geared towards a novice user:

* Moving the simplest workflow (`generate_collection_umm_var`) to be first.
* Expanding the content in the markdown cells (partially by migrating comments in the code into the markdown cells, partially by being a bit more verbose).
* Adding markdown with more details about the `Authorization` header (as I think that's one of the trickier bits of this)
* Making sure it's clear what the user should expect to see for output in a couple of instances.
* Having a bit more of a documentation-focused introduction.

CHANGELOG.md
eni-awowale: Thanks! Haha I didn't know this lol

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Add comments about activating python environment with pyenv or conda before running any pip commands

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Reframe intro to tell end-user about what they will learn from this notebook (leave information about the version so users no what version this applies too)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+    "```"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Current output format\n",
+    "Currently `earthdata-varinfo` produces Python classes that can be used within Python scripts, Jupyter notebooks or a Python Read-Eval-Print Loop (REPL).\n",
```
eni-awowale, line 82: * [ ] Remove this sentence or reword:
```
Currently `earthdata-varinfo` produces Python classes that can be used within Python scripts, Jupyter notebooks or a Python Read-Eval-Print Loop (REPL)
```

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,314 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Keys are the full variable paths, values are \n",
+    "# UMM-Var schema-compatible, JSON-serialisable dictionaries.\n",
+    "umm_var_entry = get_umm_var(var_info, var_info.get_variable('/TROPPV'))\n",
```
owenlittlejohns, line 248: Trying to summarise discussion with David here:

* This can be tricky if a user specifies the wrong variable path (e.g., a variable that doesn't exist), because that will mean the second argument is `None` and `get_umm_var` will fail.
* In practice people will most likely be iterating through the keys of variables in the `VarInfoFromNetCDF4` class, so shouldn't hit this.
* To maybe clear up confusion, though, the `var_info.get_variable('/TROPPV')` can be made it's own variable on a line above `umm_var_entry = get_umm_var(...)`, and there can be a bit of discussion about how the `VarInfoFromNetCDF4.get_variable` method can return `None` if a bad variable path is used.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: Good idea!
* [ ] Switch functions around to one function to rule them all first

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: 💀 💀 💀

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Write a markdown cell for elaborating on tokens (edl vs bearer token stuff)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: *[ ] Make comes their own markdown cells if they are long and hold important info

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Change `out_directory` to not include `/usr`

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Add example of return code

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Add truncated output

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Create cell for setting collection concept id

eni-awowale: Can we actual use two example datasets? I think it's clearer if we have one that we use `generate_collections_umm_var` for and another we use all smaller functions for

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,312 @@
...
```
autydp, line 248: Suggestion: 
TROPPV_var = var_info.get_variable('/TROPPV')
if TROPPV_var:  # best to check for None first
    umm_var_entry = get_umm_var(var_info, TROPPV_var)
else
    print("/TROPPV not found in granule")

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
autydp, line 1: Perhaps a bit less overlap with ReadMe and focus on publish to CMR.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: Yeah those were my thoughts!

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
owenlittlejohns: Overall structure looks great.

Putting the old PEP8 hat on for a nit: checks against `None` should be: `if TROPPV_var is not None`, rather than just `if TROPPV_var`.

autydp: √ - I bow to PEP8!

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: Thanks! @owenlittlejohns had a similar comment. I will add this to my next commit.
*[ ] Better overview

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Get the granule record and granule download URL with `get_granules` and `get_granule_link`\n",
+    "\n",
+    "* `get_granules`: queries `CMR_OPS` for a UMM-G record (granule record) given a collection or granule concept-id\n",
```
owenlittlejohns, line 233: Nit: Doesn't it query `CMR_UAT` rather than `CMR_OPS`? (Although totally worth saying something about "if you want to query production, change this to `CMR_OPS`...")

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "`generate_collection_umm_var` will:\n",
+    "\n",
+    "* Download the most recent granule for **GLDAS_NOAH10_3H**\n",
+    "* Generate the UMM-Var records for this granule\n",
+    "* Publish these records to CMR if `publish=True` and return a list of published variable concept-ids\n",
+    "(e.g. `['V1259971755-EEDTEST', 'V1259971757-EEDTEST' ...]`)"
```
owenlittlejohns, line 165: What about if there is an error?

Seems like a user would want to know what that looks like, and how we'd recommend fixing any issue. (Not sure if this is the best approach, but potentially using the nitty-gritty function targeted to deal with the single variable, allowing local resolution of the UMM-Var record, maybe?)

owenlittlejohns (commented):
I just had a few small nits still. (They are getting more in the weeds, so I think this is pretty close!)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: `cmr_search.get_granules` queries `CMR_OPS` by default. Here is the [line](https://github.com/nasa/earthdata-varinfo/blob/main/varinfo/cmr_search.py#L25C1-L25C1)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "generate_collection_umm_var(collection_concept_id=collection_concept_id_gldas,\n",
+    "                            auth_header=auth_header, publish=True)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Publishing and creating UMM-Var entries for **M2I1NXASM** with:\n",
```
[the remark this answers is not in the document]
eni-awowale: * [ ] Explain this is an alternative to `generate_collection_umm_var`

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
...
+    "* [GLDAS_NOAH10_3H](https://cmr.uat.earthdata.nasa.gov/search/collections.umm_json?concept-id=C1256543837-EEDTEST)\n",
+    "* [M2I1NXASM](https://cmr.uat.earthdata.nasa.gov/search/collections.umm_json?concept-id=C1256535511-EEDTEST)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Authorization:\n",
```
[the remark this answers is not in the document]
eni-awowale: I will update to this:

### Authorization:

* Launchpad or EDL tokens must be used in order query and publish to CMR.
* EDL tokens contain the header prefix `Bearer` before the token
    * For example: `Bearer <EDL token>`
* Launchpad tokens do **NOT** contain any prefixes for the header
    * For example: `<launchpad token>`
 
To request a Launchpad Token visit:
* [Launchpad Authentication User's Guide](https://wiki.earthdata.nasa.gov/display/CMR/Launchpad+Authentication+User%27s+Guide)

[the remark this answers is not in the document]
eni-awowale: Updated to suggestion on slack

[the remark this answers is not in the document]
eni-awowale: * [ ] Update this

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
owenlittlejohns: Totally fair, but in this notebook we are setting the value to `CMR_UAT` in the cell below. (Maybe the extra text should be to clarify there's a default value of `CMR_OPS`)

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## How to use `earthdata-varinfo` to publish UMM-Var records to CMR"
```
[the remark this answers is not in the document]
eni-awowale: *[ ] Why do i have half a dozen markdown cells? Combine them lol

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Other notebook requirements:\n",
+    "\n",
+    "Required packages for `earthdata-varinfo` should be installed as dependencies of PyPI. For local development, without a standard pip installation, third party requirements can be installed from the following files:\n",
```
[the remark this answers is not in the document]
eni-awowale: Honestly, I copied that line from `earthdata-varinfo` notebook. Should I just deleted this whole section lol?

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   "source": [
+    "## How to use `earthdata-varinfo` to publish UMM-Var records to CMR"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Version 2.0.0"
```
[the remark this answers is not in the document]
eni-awowale: * [ ] Remove line about version

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: Oh right we are doing that haha!

eni-awowale: Since it's just a search query I am going to remove that line in `get_granules` (`cmr_env=CMR_UAT` ) and just leave it as the default behavior which is OPS and leave something in the comments about being able to change it from the default, if that's cool?

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Add example including errors

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
eni-awowale: * [ ] Add text about being able to query different environments

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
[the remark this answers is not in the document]
owenlittlejohns: Well, looks like I should fix up the language in that other notebook. I'll fold that in to the DAS-1966 PR now.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
owenlittlejohns: I think the reason the `cmr_env=CMR_UAT` was included was to avoid someone accidentally just running all the cells and updating production metadata records, which are end-user facing. But either way works, so long as it's clear what the user has to do. (Maybe we need a `sudo`/Spiderman disclaimer at the top of the notebook about great power/responsibility?)

eni-awowale: For all of the cells with `publish_all_umm_var` and `publish_umm_var` I have the environment set to `CMR_UAT` so in the case that a user was to run through all the cells at once it won't update production metadata records

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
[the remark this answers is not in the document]
eni-awowale: * [ ] Reword

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
owenlittlejohns: Thanks for switching things around. And I really like that you have the separate `import` statement that just grabs `generate_collection_umm_var`; it highlights just how much less you need to do that the alternative workflow.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
owenlittlejohns: Nice - thanks for working on this. It's a tricky topic, and I think the current version helps out a lot.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
owenlittlejohns: Thanks for clearing this up - I think the cell as it is now makes it really clear what might happen. Nice.

docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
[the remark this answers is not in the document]
owenlittlejohns: I've also updated the other notebook.

owenlittlejohns (approved):
Thanks for fixing all the nits. I think this is looking great.

The best feedback will come from when curators try to use `earthdata-varinfo` to generate their variables, and if this is helpful - I think it will be, for sure!

PR #13, opened 2023-10-10
changed env to cmr_uat

eni-awowale:
## Description

Updated to `cmr_env` to `CMR_UAT


owenlittlejohns (approved):
Looks good; I think it's a good precaution to make the notebook run against UAT. (That way if someone just tells the notebook to run all the cells production CMR records shouldn't be affected)

PR #14, opened 2023-10-16
DAS-1099

eni-awowale:
## Description

Updated `is_science_variable()` and `get_science_variables`. `is_science_variable()` returns true if a variable is a science variable by checking if a variable contains dimension scale variables or if it contains coordinate references
or grid mapping attribute variables. `get_science_variables` returns a `set()` of variables that are science variables. Updated `umm_var.get_umm_var` to include the function `get_umm_var_type`, which adds the UMM-Var field "VariableType" if a record is a science variable

## Local Test Steps
```
from varinfo import VarInfoFromNetCDF4, VarInfoFromDmr
merra = VarInfoFromNetCDF4('../MERRA2_400.tavg1_2d_slv_Nx.20210831.nc4')
merra.get_science_variables()
```

## PR Acceptance Checklist
* ~[ ] Jira ticket acceptance criteria met.~
* ~[ ] `CHANGELOG.md` updated to include high level summary of PR changes.~ 
* ~[ ] Tests added/updated and passing.~
* [ ] Documentation updated (if needed).


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## v2.0.1
```
owenlittlejohns, line 1: I think because this adds a new public method to `VariableBase`, and adds a new fields to the output UMM-Var records, this increment should be a minor version bump to v2.1.0, not a patch. (I'm also going to claim that the changes to `get_science_variables` are a bug-fix, and not a major change.)

varinfo/variable.py
```diff
@@ -235,6 +235,25 @@ def is_temporal(self) -> bool:
         """
         return ' since ' in self.attributes.get('units', '')
 
+    def is_science(self) -> bool:
+        """ Determine if a variable is a science variable.
+            A science variable is classified as a variable that is not a
+            geographic, temporal, or a projected spatial dimension variable,
```
owenlittlejohns, line 241: I think this definition is a tiny bit wonky. I was expecting that we'd be iterating through the dimensions on the variable and checking that any of those were geographic, temporal or projected x/y. Something like:

```
if any(
    dimension.is_geographic()
    or dimension.is_temporal()
    or dimension.is_projection_x_or_y()
    for dimension in self.references.get('dimensions', [])
):
    return True
```

Also, I think the second conditional block is kind of the reverse... if a variable has `coordinates` or `grid_mapping`, I would have thought that they _are_ science variables. For example, I'm thinking of a swath science variable for the `coordinates` case.

(That would mean the default return value is `False`)

varinfo/umm_var.py
```diff
@@ -71,6 +71,10 @@ def get_umm_var(var_info: VarInfoBase, variable: VariableBase) -> Dict:
 
     """
     variable_name = variable.full_name_path.lstrip('/')
+    if variable.is_science():
```
owenlittlejohns, line 74: I'd prefer to follow the pattern of the other optional attributes for `VariableType`, and put this conditional block into it's own function, something like `get_variable_type`. That way it keeps this function a bit cleaner, and the scope consistent. Also, it will be easier to iterate that function later ~~if~~ when we find cunning ways to identify other variable types and subtypes.

tests/unit/test_umm_var.py
```diff
@@ -218,7 +218,7 @@ def test_get_umm_var_absent_fields_removed(self):
         # values have been removed from the record):
         self.assertSetEqual(
             set(umm_var_record.keys()),
-            {'Name', 'LongName', 'Definition', 'DataType',
+            {'Name', 'LongName', 'Definition', 'DataType', 'VariableType',
```
owenlittlejohns, line 221: I think this test should also update the assertion below this one that checks that the UMM-Var JSON record contains a subset matching the expected record. In this case, `precipitationCal` should also have `'VariableType': 'SCIENCE_VARIABLE'`.

tests/unit/test_umm_var.py
owenlittlejohns, line 1: Once a new method (something like `get_variable_type`) is added to this module, it should end up with its own couple of unit tests. (Just checking that you get `SCIENCE_VARIABLE` for when `variable.is_science = True`, and it is `None` when `variable.is_science = False`)

CHANGELOG.md
owenlittlejohns: Also, I reckon we could probably release these changes as part of this PR. So that would mean adding the date and then updating `VERSION`.

varinfo/variable.py
```diff
@@ -235,6 +235,25 @@ def is_temporal(self) -> bool:
...
+            attribute.
+
+        """
+        if (self.is_geographic()
+            or self.is_temporal()
+                or self.is_projection_x_or_y()):
+            return False
+
+        if (self.references.get('coordinates')
+                or self.references.get('grid_mapping')) is not None:
```
owenlittlejohns, line 252: I think this conditional is effectively doing what you want it to, but probably not the way you expect. I think what you probably mean is:

```
if (
    self.references.get('coordinates') is not None
    of self.references.get('grid_mapping') is not None
):
    ...
```

What's happening now is the parentheses are grabbing the first "truthy" value and then comparing that to `None`. The way the `references` dictionary is defined, I think this won't be a problem, but it seems cleaner to do the direct comparison to `None` for both.

owenlittlejohns (changes requested):
Nice work - I am really excited that DAS-1099 is getting taken care of (and a little bit more than the original scope)!

My biggest comments were about the `is_science` method. Also, there are no unit tests for that method in `tests/unit/test_variable.py`. (When you get to writing those, I'd recommend writing the tests around some DMR-style inputs, because then we don't need to have a bunch of NetCDF-4 test files in the repo)

Last general comment - when squash-and-merging, having a descriptive commit message would be great (e.g., "DAS-1099 - Update science variable definition." or something like that).

varinfo/umm_var.py
eni-awowale: * [ ] create `get_variable_type`

CHANGELOG.md
eni-awowale: * [ ] update changelog to 2.1.0

tests/unit/test_umm_var.py
eni-awowale: * [ ] write unit test for `get_variable_type`

varinfo/variable.py
owenlittlejohns: Updating with a summary of Slack conversation...

This isn't as easy as I'd first hoped, because the `VariableBase` class does not have access to the parent `VarInfoBase` instance. (Circular dependencies and whatnot). Right now the values in the `VariableBase.references` dictionary are sets of variable paths, not the `VariableBase` instances associated with those variables.

The alternative, to avoid circular dependencies is to move the `VariableBase.is_science` method to be a `VarInfoBase.is_science_variable` method. (This allows us to access other variables in the granule, so the dimensions can be assessed)

I think this means:

1) `VariableBase.is_science` can become `VarInfoBase.is_science_variable` (disclaimer, there maybe minor typos below...):

```
def _is_spatiotemporal_dimension(self, dimension_path: str) -> bool:
    """ A helper method to simplify the list comprehension in `is_science_variable`. """
    dimension = self.get_variable(dimension_path)
    return (
        dimension is not None
        and (dimension.is_geographic or dimension.is_temporal() or dimension.is_projection_x_or_y())
    )

def is_science_variable(self, variable: VariableBase) -> bool:
    if any(
        self._is_spatiotemporal_dimension(dimension)
        for dimension in variable.references.get('dimensions', [])
    ):
            return True

    if (
        variable.references.get('coordinates') is not None
        or variable.references.get('grid_mapping') is not None
    ):
        return True

    return False
```

2) The `VarInfoBase.get_science_variables` method needs to update its set comprehension to:

```
filtered_with_coordinates = {
    variable_path
    for variable_path, variable
    in self.variables.items()
    if variable_path is not None
    and self.is_science_variable(variable)
    and not self.variable_is_excluded(variable_path,
                                      exclusions_pattern)
}
```

3) The `VarInfoBase.get_metadata_variables` method needs to update its set comprehension to:

```
non_coordinate_variables = {
    variable_path
    for variable_path, variable
    in self.variables.items()
    if variable_path is not None
    and (self.variable_is_excluded(variable_path, exclusions_pattern)
         and not self.is_science_variable(variable))
}
```

(Sorry for the essay!)

eni-awowale: No sorry's! I needed the essay!!!

tests/unit/test_umm_var.py
eni-awowale: * [ ] Add `'VariableType': 'SCIENCE_VARIABLE'`

eni-awowale: FYI this variable `/precipitationUncal` is not a science variable. So I made a different test for a known science variable. I will also make another test to make sure dimension variables aren't getting marked as science variables.

eni-awowale: I updated `test_get_umm_var_all_fields` instead since that test has dimension variables assigned to science variables

varinfo/variable.py
eni-awowale: * [ ] updated

tests/unit/test_var_info.py
```diff
@@ -341,9 +344,7 @@ def test_var_info_get_metadata_variables(self):
                                  config_file=self.config_file)
 
         metadata_variables = dataset.get_metadata_variables()
-        self.assertEqual(metadata_variables,
-                         {'/required_group/has_no_coordinates',
-                          '/exclude_one/has_coordinates', '/science/lat_bnds'})
+        self.assertSetEqual(metadata_variables, set())
```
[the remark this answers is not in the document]
flamingbear: ok, This looks like you changed it on purpose, but I don't know if that's what you really wanted to do. Can you double check?

varinfo/var_info.py
```diff
@@ -189,6 +189,38 @@ def get_variables_with_coordinates(self) -> Dict[str, OutputVariableType]:
...
+            if dimension != variable.full_name_path
+            and not variable.full_name_path.endswith('_bnds')
+        ):
+            return True
+
+        if (
+            variable.references.get('coordinates') is not None
+                or variable.references.get('grid_mapping')) is not None:
+            return True
+
```
flamingbear, line 221: That is almost certainly not what you intended if you are trying to make sure that neither `coordinates` nor `grid_mapping`  are None.   

If this works, I think this works by accident, you probably want to move the parenthesis for readability.

varinfo/var_info.py
```diff
@@ -189,6 +189,38 @@ def get_variables_with_coordinates(self) -> Dict[str, OutputVariableType]:
...
                 and not self.variable_is_excluded(variable, exclusion_pattern)}
 
+    def _is_spatial_temporal_dimension(self, dimension_path: str) -> bool:
+        """ A helper method simplifying the list comprehension in
+            `is_science_variable`.
+        """
+        dimension = self.get_variable(dimension_path)
+        return (dimension is not None
+                and (dimension.is_geographic() or dimension.is_temporal()
+                     or dimension.is_projection_x_or_y()))
```
flamingbear, line 199: This is simply for readability.  This is easier to read to me, but total nit.
```suggestion
        return dimension is not None and (
            dimension.is_geographic()
            or dimension.is_temporal()
            or dimension.is_projection_x_or_y()
        )
```

flamingbear (changes requested):
We should tag up face to face tomorrow if possible because I'm not super familiar with this, but some of these changes don't make sense to me.

varinfo/variable.py
flamingbear: And here is where I think that error was introduced. :stare:

varinfo/var_info.py
```diff
@@ -231,7 +264,7 @@ def get_metadata_variables(self) -> Set[str]:
             in self.variables.items()
             if variable_path is not None
             and (self.variable_is_excluded(variable_path, exclusions_pattern)
-                 or variable.references.get('coordinates') is None)
+                 and not self.is_science_variable(variable))
```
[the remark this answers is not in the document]
flamingbear: @eni-awowale and I worked through this and the updates are in f4bdbfd

tests/unit/test_umm_var.py
```diff
@@ -950,3 +950,18 @@ def test_generate_variable_native_id(self):
...
+            a science variable.
+        """
+        science_variable_type = get_variable_type(self.merra_varinfo,
+            self.merra_varinfo.variables.get('/EPV'))
+
+        not_science_variable_type = get_variable_type(self.merra_varinfo,
+            self.merra_varinfo.variables.get('/time'))
+
+        self.assertEqual('SCIENCE_VARIABLE', science_variable_type)
+        self.assertFalse(not_science_variable_type)
```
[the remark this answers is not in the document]
eni-awowale: * [ ] change to `self.assertIsNone()`

varinfo/var_info.py
eni-awowale: added

flamingbear (approved):

flamingbear:
Thanks for cleaning up those nits and walking me through the other stuff.  That was really helpful, I think this looks good.

eni-awowale:
Thanks @flamingbear for going through this with me!

PR #16, opened 2023-11-15
Use CF-Convention long_name in UMM-Var.

owenlittlejohns:
## Description

This PR updates the generation of the `LongName` attribute in UMM-Var records to preferentially use the `long_name` CF-Convention attribute if it is present for a variable.

Due to this change of definition in `LongName`, the native ID for the variable has been changed from `<collection concept ID>-<variable LongName>` to `<collection concept ID>-<variable Name>`. This change will have no impact on previously generated records as the `Name` and `LongName` fields generated were previously identical.

## Local Test Steps

* Pull this version of the code and install it with `pip install -e .`
* Run the UMM-Var generation routine on a collection with variables that have a `long_name` attribute. Check that value is used in place of the variable path.

## PR Acceptance Checklist


.github/workflows/run_tests.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.9', '3.10', '3.11', '3.x' ]
+        python-version: [ '3.9', '3.10', '3.11' ]
```
owenlittlejohns, line 11: This is a temporary measure. `distutils` was deprecated and is no longer available in Python 3.12 (which was released recently). Somewhere in our dependency tree, something is trying to install `distutils`, so this is  causing issues. I think resolving that (requiring some sleuthing on our dependency tree) is outside the scope of this PR. Will address separately.

CHANGELOG.md
```diff
@@ -1,9 +1,20 @@
+## v2.1.2
+### 2023-11-14
+
+This version of `earthdata-varinfo` updates the value of the `LongName`
+attribute in generated UMM-Var records to use the value of the CF-Convention
+`long_name` attribute for a variable, if it is present in the file. If this
+attribute is not present in the in-file metadata, then the full path to the
```
eni-awowale, line 7: Might be a super nit:
Is it the full path name that was used before? I thought the full path usually started with a `/`

tests/unit/test_umm_var.py
```diff
@@ -188,6 +188,44 @@ def test_get_umm_var_all_fields(self):
...
+        nc_varinfo = VarInfoFromNetCDF4(netcdf4_file)
+
+        with self.subTest('CF-Convention long_name is used'):
+            long_name_umm_var = get_umm_var(
+                nc_varinfo, nc_varinfo.get_variable('/long_name')
+            )
+
+            self.assertEqual(
+                long_name_umm_var['LongName'],
+                'this is really long'
```
eni-awowale, line 216: `'this is really long'`
LOL

eni-awowale:
Looks good! I had one super nit related to more semantics and language but everything is working as expected!

eni-awowale (approved):

CHANGELOG.md
owenlittlejohns: The default is still the value that was used before, which is the full path (stripped of the leading `/` to conform with UMM-Var expectations). I'll clarify this comment and then merge. Good catch.

PR #21, opened 2024-04-11
IP-241 - Add pre-commit CI/CD.

owenlittlejohns:
## Description

This PR is to add [pre-commit.ci](https://pre-commit.ci/) to `earthdata-varinfo`, which will automatically run our `pre-commit` hooks for each PR. I've configured it to _not_ automatically add a new commit with the changes, because it feels better to still have a human in the loop to review those changes.

Steps I did to set this up:

1) Log in to [pre-commit.ci](pre-commit.ci).
2) Select repositories for which I (or the person doing this) is an administrator, and add them to `pre-commit.ci`.
3) Update the `.pre-commit-config.yaml` to add the `ci` section.
4) Feel pleased about life.

## Jira Issue ID

N/A

## Local Test Steps

N/A

## PR Acceptance Checklist
* ~~`docker/service_version.txt` updated if publishing a release.~~
* ~~Tests added/updated and passing.~~

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,5 @@
+ci:
+  autofix_prs: false
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v3.2.0
--- a/README.md
+++ b/README.md
@@ -239,6 +239,9 @@ review and `git add` the changes before you can make a commit.
 It is planned to implement additional hooks, possibly including tools such as
 `mypy`.
 
+[pre-commit.ci](pre-commit.ci) is configured such that these same hooks will be
+automatically run for every pull request.
+
 ## Releasing:
 
 All CI/CD for this repository is defined in the `.github/workflows` directory:


flamingbear (approved):
All the things!

eni-awowale:
@owenlittlejohns I don't have administrative privileges to this repository so I can't update this

.pre-commit-config.yaml
eni-awowale, line 1: Something I have run into when using `.pre-commit` and setting the language version to be explicitly `3.11`; the `ruff` and `black` hooks don't like `TypeHints` that are imported from `typing` as `Dict`, `List`, or when you do `Union[str, int]`  as a return type. But if you import them as just `dict`, `list` and do `str|int` for a return type it breaks on 3.9 or lower older. Just something to note

.pre-commit-config.yaml
```diff
@@ -1,3 +1,5 @@
+ci:
+  autofix_prs: false
```
eni-awowale, line 2: Does this turn off auto formatting?

flamingbear: turns off auto committing the changes.  This will make the format changes and require you to add them to your commits.

eni-awowale: Ah okay I like that too!

flamingbear: The thing Owen did before actually does what I said, this is for Automatic github CI.  And instead of fixing it for you with a new commit, it will fail the PR checks.  You will have to push up the changes to fix it yourself.

owenlittlejohns: Yeah - we can discuss this some more, but I get pretty squeamish at the idea of getting those commits added for free. Aside from that, it's probably more useful for developers to understand why the check failed, so they remember to not do the same thing in future work.

.pre-commit-config.yaml
owenlittlejohns: Fair point. That can be taken care of with `__future__`, I think. (That's what has been done in `xarray` at least)

owenlittlejohns:
> @owenlittlejohns I don't have administrative privileges to this repository so I can't update this

@eni-awowale - The steps outlined in the PR description are the one-time set-up steps that had to be done to get this to work. So Step 2 is something that only needs to be done once by a repository admin. I mainly wanted to make sure that the information wasn't lost and that people knew the steps I did (for future reference, or repeating the same steps on a different repo).

.pre-commit-config.yaml
eni-awowale: What I was getting at initially is that I prefer turning off auto-formatting (personal preference), except for small things like white spaces  or `' '` vs `" "` . I like to see where my code fails and then change it and re-commit it myself. This seems to solve that problem by not automatically re-committing so that works too!

eni-awowale:
> > @owenlittlejohns I don't have administrative privileges to this repository so I can't update this
> 
> @eni-awowale - The steps outlined in the PR description are the one-time set-up steps that had to be done to get this to work. So Step 2 is something that only needs to be done once by a repository admin. I mainly wanted to make sure that the information wasn't lost and that people knew the steps I did (for future reference, or repeating the same steps on a different repo).
> 
> Edit: One thing I left out is that you can see this worked by looking at the checks that were run. One of them now is the pre-commit one. 🎉

Okay awesome!! I thought that was something I needed to do too!

eni-awowale (approved):
Yay `.pre-commit`!

PR #28, opened 2024-07-15
DAS-2165 - Add a config file parameter to generate_collection_umm_var method

sudha-murthy:
## Description

- An optional config file parameter added to VarInfoFromNetCDF4
- Unit tests added to verify that the argument is passed in 
- Updates to Workflow 1 in how_to_publish_to_cmr_with_earthdata_varinfo.ipynb to include the config_file optional argument

## Jira Issue ID
- DAS-2165

## Local Test Steps
- Use the README.md to setup the local conda environment
- Run make develop
- Run make test
- All tests should pass

## PR Acceptance Checklist
* [X] Jira ticket acceptance criteria met.
* [X] `CHANGELOG.md` updated to include high level summary of PR changes.
* [X] `VERSION` updated if publishing a release.
* [X] Tests added/updated and passing.
* [X] Documentation updated (if needed).


docs/how_to_publish_to_cmr_with_earthdata_varinfo.ipynb
```diff
@@ -251,6 +253,27 @@
     ")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Instantiate a VarInfoFromNetCDF4 object with an optional config file. This will override any default configuration."
```
flamingbear, line 260:
```suggestion
    "Instantiate a VarInfoFromNetCDF4 object with an optional config file. This will override default configuration."
```
I don't think having a config fill will override Any defaults.  This just sounds better to me.

tests/unit/test_generate_umm_var.py
```diff
@@ -294,3 +294,78 @@ def test_is_variable_concept_id(self):
...
+        mock_granule_query.return_value.get.return_value = [
+            {
+                'links': [
+                    {
+                        'href': self.netcdf4_url,
+                        'rel': 'http://esipfed.org/ns/fedsearch/1.1/data#',
+                    }
+                ]
+            }
+        ]
```
flamingbear, line 319: This code snippet is repeated 4 times in this file.  I would make this particular value a class variable and just reference it each time you need to use it.

tests/unit/test_generate_umm_var.py
```diff
@@ -294,3 +294,78 @@ def test_is_variable_concept_id(self):
...
+    @patch('varinfo.generate_umm_var.download_granule')
+    def test_generate_collection_umm_var_config_file(
+        self, mock_download_granule, mock_granule_query, mock_varinfo_from_netcdf4
+    ):
+        """A request with all the necessary information should succeed.
+        This test is moderately end-to-end, but mocks the full
+        download_granule function for simplicity. That function is tested
+        in detail in test_cmr_search.py.
+
+        """
```
flamingbear, line 309: Probably update this with information for this test.

tests/unit/test_generate_umm_var.py
```diff
@@ -294,3 +294,78 @@ def test_is_variable_concept_id(self):
...
+    @patch('varinfo.cmr_search.GranuleQuery')
+    @patch('varinfo.generate_umm_var.download_granule')
+    def test_generate_collection_umm_var_with_no_config_file(
+        self, mock_download_granule, mock_granule_query, mock_varinfo_from_netcdf4
+    ):
+        """A request with all the necessary information should succeed.
+        This test is moderately end-to-end, but mocks the full
+        download_granule function for simplicity. That function is tested
+        in detail in test_cmr_search.py.
+
```
flamingbear, line 348: Update this with correct information

varinfo/generate_umm_var.py
```diff
@@ -70,7 +72,7 @@ def generate_collection_umm_var(
         )
 
         # Parse the granule with VarInfo to map all variables and relations:
-        var_info = VarInfoFromNetCDF4(local_granule)
+        var_info = VarInfoFromNetCDF4(file_path=local_granule, config_file=config_file)
```
flamingbear, line 75: Why did you make `file_path` a keyword argument?  I would leave it. 
```suggestion
        var_info = VarInfoFromNetCDF4(local_granule, config_file=config_file)
```
Isn't that the same?

flamingbear (changes requested):
This looks mostly good.  I did ask for some small refactoring of the test fixtures and had a couple of nit-pick rewords.  Nice job.

varinfo/generate_umm_var.py
sudha-murthy: Did not realize I made the change. will undo the change if I did

CHANGELOG.md
```diff
@@ -1,13 +1,11 @@
 
 ## v2.2.2
-### Unreleased
```
flamingbear: Sorry, you should leave this one in.  But are you going to release it?  If so this should have a date 

```
### 2024-07-16
```

This is probably worth releasing, but I'm not sure who makes that call.

tests/unit/test_generate_umm_var.py
```diff
@@ -19,6 +19,7 @@ def setUpClass(cls):
         cls.collection_concept_id = 'C1234567890-PROV'
         cls.launchpad_token_header = 'launchpad-token'
         cls.netcdf4_basename = 'f16_ssmis_20210426v7.nc'
+        cls.data_url = 'http://esipfed.org/ns/fedsearch/1.1/data#'
```
flamingbear, line 22: Sorry, I think you misunderstood what I was trying to say:

For the class var set it up for the full value
``` python
cls.query_granule_return = [
            {
                'links': [
                    {
                        'href': self.netcdf4_url,
                        'rel': 'http://esipfed.org/ns/fedsearch/1.1/data#',
                    }
                ]
            }
        ]
```

And then you can use it everywhere below
``` python
        mock_granule_query.return_value.get.return_value = self.query_granule_return
```

tests/unit/test_generate_umm_var.py
flamingbear: It doesn't look like you updated this comment?

tests/unit/test_generate_umm_var.py
flamingbear: And this comment is still the same.

flamingbear (changes requested):
Sorry I wasn't clear in my first comments, hopefully this makes more sense.

tests/unit/test_generate_umm_var.py
sudha-murthy: I forgot to commit that change. done now

tests/unit/test_generate_umm_var.py
sudha-murthy: ok. makes sense.

CHANGELOG.md
sudha-murthy: Not sure ...

flamingbear (approved):
Thanks for the changes.  I think this looks good to go. :shipit:

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
+
+## v2.2.2
```
owenlittlejohns, line 2: Nitpick: `generate_collection_umm_var` is one of the end-user entry points of our code, so I think this change counts as a backwards compatible API change. I think that means this should be a minor version number change (so 2.3.0 rather than 2.2.2).

varinfo/generate_umm_var.py
```diff
@@ -37,6 +37,7 @@ def generate_collection_umm_var(
     auth_header: str,
     cmr_env: CmrEnvType = CMR_UAT,
     publish: bool = False,
+    config_file: Optional[str] = None,
```
owenlittlejohns, line 40: A preferred alternative here would be:

```
config_file: str | None = None
```

Because we also support Python 3.9, though, that means you need to add:

```
from __future__ import annotations
```

this goes right near the top of the file.

owenlittlejohns (commented):
I was just reviewing this as you hit merge (sorry for being out today).

I only had two comments, nothing major, and I'll tweak the type hint one (the `str | None` instead of `Optional` in my next PR for the configuration file schema clean-up).

Nice work.

varinfo/generate_umm_var.py
sudha-murthy: Thanks Owen for the comments

sudha-murthy: I switched that change 
    config_file: str | None = None 
back since it was failing for 3.9
Does the __future__ import basically make it not 3.9 anymore?

CHANGELOG.md
sudha-murthy: ok

varinfo/generate_umm_var.py
owenlittlejohns: The import of `__future__.annotations` will let the Python compiler use some specific newer features than that version had access to when it was released. [Here are the docs for the future part of the standard library](https://docs.python.org/3/library/__future__.html).

PR #35, opened 2024-08-26
TRT-290 - Augment documentation notebook regarding required variables.

owenlittlejohns:
## Description

This PR updates the main `earthdata-varinfo` documentation Jupyter notebook to clarify the CF Convention metadata attributes that are used to determine "required variables".

## Jira Issue ID

TRT-290 (this request came up in the context of supporting SMAP migration to the cloud, but the Jira issue is a bit tangential).

## Local Test Steps

N/A

## PR Acceptance Checklist
* ~~Tests added/updated and passing.~~


docs/earthdata-varinfo.ipynb
```diff
@@ -198,6 +198,19 @@
    "source": [
     "In the output above, `/Grid/precipiationCal` has three dimensions: `/Grid/time`, `/Grid/lat` and `/Grid/lon`. These dimension variables also refer to their respective [bounds variables](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#cell-boundaries). Because `get_required_variables` is recursive, the bounds attributes are also considered required variables for `/Grid/precipitationCal`.\n",
     "\n",
+    "`earthdata-varinfo` checks for references in known CF Convention metadata attributes that are expected to contain references to other variables. These include:\n",
```
flamingbear, line 201: Is it earthdata-varinfo checking for references or this particular call to `get_required_variables`?

docs/earthdata-varinfo.ipynb
```diff
@@ -198,6 +198,19 @@
...
+    "\n",
+    "* [ancillary_variables](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#ancillary-data)\n",
+    "* [coordinates](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#coordinate-types), see also [here](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#coordinate-system).\n",
+    "* [grid_mapping](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#grid-mappings-and-projections)\n",
+    "* [bounds](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#cell-boundaries)\n",
+    "* [cell_measures](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#cell-measures)\n",
+    "* [geometry, interior_ring, node_coordinates, node_count, nodes and part_node_count](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.11/cf-conventions.html#geometries)\n",
+    "\n",
+    "The check is recursive, as a variable referred to in the initial set specified in the call to the `get_required_variables` method might itself refer to other variables.\n",
```
flamingbear, line 210: You do mention it's recursive right above this change also, I don't think it's a big deal, but wasn't sure if you wanted to reword a bit?  Maybe `Because the check is recursive,` would be enough

flamingbear (approved):
checked all the links, they go where I would expect. take or leave my questions/advice.

docs/earthdata-varinfo.ipynb
owenlittlejohns: This is a good question. Here's the long answer:

The `VarInfo*` (or perhaps more specifically: `Variable*`) classes will try to identify all the references when a file is initially parsed. But all that means is checking the expected metadata attributes and then trying to fully qualify the references it finds based on the location of the variable making that reference and the value of the metadata attribute. (That is - there is no checking that the variable being referred to exists, and there is no recursive tracing at that point)

`VarInfoBase.get_required_variables` is where the dependencies are traced to get a full set.

So maybe what I'm saying is it's a little of column A and a little of column B.

**TL;DR: Maybe a more accurate update would be:**

> When a `VarInfoFromDmr` or `VarInfoFromNetCDF4` object is instantiated, specific CF Convention metadata attributes of each variable are checked for references references to other variables. These references are then fully qualified to absolute paths for later use by methods such as `VarInfoNetCDF4.get_required_variables`. The metadata attributes expected to contain references to other variables, per the CF Conventions include:

docs/earthdata-varinfo.ipynb
owenlittlejohns: I'll make that change and push it up with the potential other change being discussed, once we settle on wording there.

owenlittlejohns: Looking at it, and the content from before, is this sentence actually adding anything? Could we just cut it in favour of the comment above about how the `bounds` variables are included.

flamingbear: I don't know that was my first thought but then again people skim documentation so repeating isn't terrible.

docs/earthdata-varinfo.ipynb
flamingbear: I think this more accurate update is better.

owenlittlejohns: Cool. I have added that here: [b50c3d](https://github.com/nasa/earthdata-varinfo/pull/35/commits/b50c3da3c9d2e4ecb51a76f4003b0100e9cd34fd)

docs/earthdata-varinfo.ipynb
owenlittlejohns: I couldn't think of a good way to make two entirely distinct points here, so I've cut it in [b50c3d](https://github.com/nasa/earthdata-varinfo/pull/35/commits/b50c3da3c9d2e4ecb51a76f4003b0100e9cd34fd).

PR #36, opened 2024-08-27
TRT-555 remove CF_Supplements

owenlittlejohns:
## Description

This PR fulfils the next step in simplifying the `earthdata-varinfo` configuration file schema by removing the `CF_Supplements` section, in favour of all rules there being specified as `CF_Overrides`. This is Proposal 3 in [the list of proposed schema changes](https://wiki.earthdata.nasa.gov/display/SITC/earthdata-varinfo+configuration+schema+v2.0.0+proposal). The linked wiki page also includes details on specific rules in use right now. At the time of compilation, there were no supplements that could not be expressed as overrides.

Two, more in the weeds changes:

* There's a bit of a change in the `CFConfig` public methods and attributes. I'll highlight that in the appropriate place.
* I noticed there wasn't any way in which precedence was given to conflicting overrides that specify values for the same metadata attributes. This PR implements logic to ensure the most specific override for a given variable takes precedence. (The proxy for specificity is the length of the `Variable_Pattern`)

## Jira Issue ID

TRT-555

## Local Test Steps

* Pull this branch locally.
* Run the tests - they should all pass.
* Bonus 1: If you have a file you like to parse and a configuration file, then update the configuration file to the new schema and instantiate a `VarInfoFromNetCDF4` object to check functionality still behaves as expected. If you don't have a local configuration file, then you can grab the `tests/unit/data/test_config.json` and start with that.
* Bonus 2: If you want to test this specific functionality, then amend your local configuration file to have a couple of rules for the same attribute (different values and different levels of specificity in the `Variable_Pattern` applicability). When retrieving the attributes for a variable parsed by `VarInfoFromNetCDF4`, the more specific attribute override should be used.

Example content for a configuration file (update values for mission, short name, variable pattern, etc):

```
{
  ...
  "CF_Overrides": [
    {
      "Applicability": {
        "Mission": "FakeSat",
        "ShortNamePath": "FAKE99",
        "Variable_Pattern": "/.*"
      },
      "Attributes": [
        {
          "Name": "attribute_name",
          "Value": "value_being_ignored"
        }
      ]
    },
    {
      "Applicability": {
        "Mission": "FakeSat",
        "ShortNamePath": "FAKE99",
        "Variable_Pattern": "/group/variable"
      },
       "Attributes": [
        {
          "Name": "attribute_name",
          "Value": "value_taking_precedence"
        }
      ]
    }
  ]
}
```

## PR Acceptance Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ an input file.
 * `CFConfig.get_cf_attributes` has been renamed `CFConfig.get_cf_overrides`, as
   there are now only overrides to be returned from this method.
   `CFConfig.get_cf_overrides` now _must_ specify a variable path. All overrides
-  from a configuration file for a given collection is now retrievable from
+  from a configuration file for a given collection are now retrievable from
   the newly public `CFConfig.cf_overrides` class attribute.
 
 ### Removed:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,22 @@ an input file.
 * Handling of nested `Applicability_Groups` has been removed from the `CFConfig`
   class, as the configuration file no longer nests these items in overrides
   or supplements.
+* CF Overrides retrieved for a matching file path are ordered such that the
+  most specific applicable override to the variable takes precedence. For
+  example, when requesting the value of the "units" metadata attribute for
+  variable "/nested/variable", an applicability rule that exactly matches this
+  variable path will take precedence over rules matching to either the group,
+  or all variables in the file.
+* `CFConfig.get_cf_attributes` has been renamed `CFConfig.get_cf_overrides`, as
+  there are now only overrides to be returned from this method.
+  `CFConfig.get_cf_overrides` now _must_ specify a variable path. All overrides
+  from a configuration file for a given collection is now retrievable from
+  the newly public `CFConfig.cf_overrides` class attribute.
+
+### Removed:
+
+* `CFConfig._cf_supplements` has been deprecated in favour of specifying all
+  in-file metadata changes via a CF override instead.
 
 ## v2.3.0
 ### 2024-08-26
--- a/config/1.0.0/earthdata_varinfo_configuration_schema.json
+++ b/config/1.0.0/earthdata_varinfo_configuration_schema.json
@@ -298,11 +298,11 @@
           "type": "string"
         },
         "Applicability": {
-          "description": "An applicability rule that can be used to define global variables, if only a mission and/or collection short name is specified, or set variable and group level metadata attributes if a Variable_Pattern is provided.",
+          "description": "An applicability rule that indicates which groups and variables within a file a metadata override should apply to. If only a short name and/or mission is provided, the override will apply to all groups and variables. If a Variable_Pattern is also provided, the override is applied only to those groups or variables whose paths match the regular expression of the Variable_Pattern.",
           "$ref": "#/$defs/ApplicabilityType"
         },
         "Attributes" : {
-          "description": "Metadata attributes to overwrite against variables or groups matching a variable path pattern specified in Applicability of this object.",
+          "description": "Metadata attributes to override for variables or groups that match the mission, short name and/or Variable_Pattern criteria specified in the Applicability of this object.",
           "type": "array",
           "items": {
             "description": "A list of metadata attributes with their names and values.",
--- a/config/1.0.0/earthdata_varinfo_configuration_schema.json
+++ b/config/1.0.0/earthdata_varinfo_configuration_schema.json
@@ -30,7 +30,7 @@
       }
     },
     "Excluded_Science_Variables": {
-      "description": "If CF_Overrides are not enough VarInfo classes will assume certain variables are science variables. This setting provides variable path patterns that should be excluded from consideration as a science variable.",
+      "description": "VarInfo classes currently assume that any variable that has a grid mapping attribute, or has a spatial or temporal dimension and is not itself a dimension or bounds variable, should be treated as a science variable. This may not be true in all cases, and so Excluded_Science_Variables provide a method to denote non-science variables that might otherwise be incorrectly identified.",
       "type": "array",
       "items": {
         "$ref": "#/$defs/MissionVariablePatternType"
@@ -232,7 +232,7 @@
           "type": "string"
         },
         "Value": {
-          "description": "The metadata attribute value. This will replace any value present.",
+          "description": "The overriding metadata attribute value. The value specified in the configuration file will replace the corresponding metadata value in any applicable source file.",
           "anyOf": [{
             "type": ["number", "string"]
           }, {
--- a/config/1.0.0/earthdata_varinfo_configuration_schema.json
+++ b/config/1.0.0/earthdata_varinfo_configuration_schema.json
@@ -294,7 +294,7 @@
       "type": "object",
       "properties": {
         "_Description": {
-          "description": "A description of the need for and effect of the overrides.",
+          "description": "Explains the purpose and effect of these overrides.",
           "type": "string"
         },
         "Applicability": {
--- a/config/1.0.0/earthdata_varinfo_configuration_schema.json
+++ b/config/1.0.0/earthdata_varinfo_configuration_schema.json
@@ -30,7 +30,7 @@
       }
     },
     "Excluded_Science_Variables": {
-      "description": "If CF_Overrides and CF_Supplements are not enough VarInfo classes will assume certain variables are science variables. This setting provides variable path patterns that should be excluded from consideration as a science variable.",
+      "description": "If CF_Overrides are not enough VarInfo classes will assume certain variables are science variables. This setting provides variable path patterns that should be excluded from consideration as a science variable.",
       "type": "array",
       "items": {
         "$ref": "#/$defs/MissionVariablePatternType"
@@ -193,14 +193,7 @@
       "description": "# For cases where CF references do not exist, or are invalid. For example, variables that have no dimension references in the HDF-5 file contents",
       "type": "array",
       "items": {
-        "$ref": "#/$defs/CFOverridesAndSupplementsItemType"
-      }
-    },
-    "CF_Supplements": {
-      "description": "For extending CF references. This allows for CF references to persist, while supplemental information is provided. In particular, e.g., for subsetting support, non-CF attributes can be provided.",
-      "type": "array",
-      "items": {
-        "$ref": "#/$defs/CFOverridesAndSupplementsItemType"
+        "$ref": "#/$defs/CFOverridesItemType"
       }
     }
   },
@@ -239,7 +232,7 @@
           "type": "string"
         },
         "Value": {
-          "description": "The metadata attribute value. For CFOverrides, this will replace any value present, for CFSupplements, this is expected to be a string and it will be appended to the present value using string concatenation.",
+          "description": "The metadata attribute value. This will replace any value present.",
           "anyOf": [{
             "type": ["number", "string"]
           }, {
@@ -296,20 +289,20 @@
       "required": ["Applicability", "Variable_Pattern"],
       "additionalProperties": false
     },
-    "CFOverridesAndSupplementsItemType": {
-      "description": "An item that details a related group of CF-Convention metadata attributes to supplement or overwrite according to the supplied applicability rules.",
+    "CFOverridesItemType": {
+      "description": "An item that details one or more metadata attributes to overwrite according to the supplied applicability rules.",
       "type": "object",
       "properties": {
         "_Description": {
-          "description": "A description of the need for and effect of the CF Convention overrides or supplements.",
+          "description": "A description of the need for and effect of the overrides.",
           "type": "string"
         },
         "Applicability": {
-          "description": "An applicability rule that can be used to define global variables, if only a mission and/or collection short name is specified, or set variable metadata attributes if a Variable_Pattern is provided.",
+          "description": "An applicability rule that can be used to define global variables, if only a mission and/or collection short name is specified, or set variable and group level metadata attributes if a Variable_Pattern is provided.",
           "$ref": "#/$defs/ApplicabilityType"
         },
         "Attributes" : {
-          "description": "Metadata attributes to supplement or overwrite against variables matching a variable path pattern specified in Applicability of this object.",
+          "description": "Metadata attributes to overwrite against variables or groups matching a variable path pattern specified in Applicability of this object.",
           "type": "array",
           "items": {
             "description": "A list of metadata attributes with their names and values.",
--- a/config/1.0.0/sample_config_1.0.0.json
+++ b/config/1.0.0/sample_config_1.0.0.json
@@ -140,6 +140,19 @@
       "Applicability": {
         "Mission": "SMAP",
         "ShortNamePath": "SPL3SMP_E",
+        "Variable_Pattern": "/$"
+      },
+      "Attributes": [
+        {
+          "Name": "Data_Organization",
+          "Value": "h5_grid"
+        }
+      ]
+    },
+    {
+      "Applicability": {
+        "Mission": "SMAP",
+        "ShortNamePath": "SPL3SMP_E",
         "Variable_Pattern": "/Soil_Moisture_Retrieval_Data_AM/.*"
       },
       "Attributes": [
@@ -178,9 +191,7 @@
           "Value": "../../grid_x ../../grid_y"
         }
       ]
-    }
-  ],
-  "CF_Supplements": [
+    },
     {
       "Applicability": {
         "Mission": "ICESat2",
@@ -346,19 +357,6 @@
         {
           "Name": "coordinates",
           "Value": "/crs"
-        }
-      ]
-    },
-    {
-      "Applicability": {
-        "Mission": "SMAP",
-        "ShortNamePath": "SPL3SMP_E",
-        "Variable_Pattern": "/$"
-      },
-      "Attributes": [
-        {
-          "Name": "Data_Organization",
-          "Value": "h5_grid"
         }
       ]
     },
--- a/tests/unit/data/test_config.json
+++ b/tests/unit/data/test_config.json
@@ -208,6 +208,58 @@
           "Value": "applies to /group/variable"
         }
       ]
+    },
+    {
+      "Applicability": {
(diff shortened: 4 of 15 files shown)


config/1.0.0/sample_config_1.0.0.json
```diff
@@ -136,6 +136,19 @@
         }
       ]
     },
+    {
```
owenlittlejohns, line 139: I've basically migrated all the supplements to overrides. The reason this one shows up in particular is that I moved it to be closer to the other SMAP L3 overrides, so the location in the file also changed.

tests/unit/data/test_config.json
owenlittlejohns, line 1: The diff on this file is conflating two things:

1) I removed all the fake supplements that were used to test that supplements were used correctly.
2) I added a couple of extra overrides to be used to test the precedence functionality I've added (e.g., for multiple matching rules that want to update the same metadata attribute, the more specific variable pattern is used).

Anyway, because they look similar ish, git has tried to be clever with the diff, and I think that maybe made it more confusing.

tests/unit/test_cf_config.py
```diff
@@ -103,27 +89,12 @@ def test_instantiation_invalid_configuration_file_format(self):
                 config_file='tests/unit/data/ATL03_example.dmr',
             )
 
-    def test_get_cf_attributes_all(self):
```
owenlittlejohns, line 106: This test has been deprecated, because of the change in behaviour for `CFConfig.get_cf_attributes` (there is no longer a branch that returns all CF Overrides for a `CFConfig` object. I'll talk more about that in place...

tests/unit/test_variable.py
```diff
@@ -231,28 +229,10 @@ def test_variable_cf_override_non_reference(self):
             variable.attributes.get('collection_override'), 'collection value'
         )
 
-    def test_variable_cf_supplement_non_reference(self):
```
owenlittlejohns, line 234: No more supplements, so no more supplement tests!

varinfo/cf_config.py
```diff
@@ -57,8 +57,7 @@ def __init__(
         self.mission = mission
         self.short_name = collection_short_name
 
-        self._cf_overrides: dict = {}
-        self._cf_supplements: dict = {}
+        self.cf_overrides: dict[str, dict[str, Any]] = {}
```
owenlittlejohns, line 60: Note - this is now a public class attribute. See the comment on `CFConfig.get_cf_overrides` for more information.

varinfo/cf_config.py
```diff
@@ -160,40 +155,43 @@ def _create_attributes_object(cf_item: dict) -> dict[str, str]:
             for attribute in cf_item.get('Attributes', {})
         }
 
-    def get_cf_attributes(
-        self, variable: str | None = None
-    ) -> dict[str, dict[str, str]]:
-        """Return the CF overrides and supplements that match a given
-        variable. If a variable is not specified, then return all overrides
-        and supplements. If there are no overrides or supplements, then
-        empty dictionaries will be returned instead.
```
owenlittlejohns, line 169: Okay, so there are two big changes to this method...

1) The variable path is now a required argument. That means that the output of this method will always be the applicable CF overrides for the named variable. Previously, if the variable path was not specified, all CF overrides for the `CFConfig` object were returned. This made the output inconsistent:

For a variable specific call:

```
{
    'attribute_name': 'attribute_value',
    'other_attribute_name': 'other_attribute_value',
}
```

For an unspecified variable call:

```
{
    '/variable/pattern': {
        'attribute_name': 'attribute_value',
        'other_attribute_name': 'other_attribute_value',
    },
    '/other/variable/pattern': {
        'attribute_name': 'attribute_value',
        'other_attribute_name': 'other_attribute_value',
    },
}
```

Inconsistent output is tricky to code around, and the full dictionary of all CF overrides is available just from the `CFConfig.cf_overrides` class attribute instead.

owenlittlejohns, line 169: 2) Precedence of conflicting rules. This is a sneaky bug that I only spotted this time around. If you have a variable name of "/this/is/my/variable" and had the following overrides:

```
{
    '/this/is/my/variable': {
        'attribute_name': 'more_specific_match',
    },
    '/this/is/.*': {
        'attribute_name': 'more_generic_match',
    }
}
```
Because the code used to just iterate through the rules and use the `dict.update` method, the last matching rule for a given attribute would take precedence. The ordering of the configuration file rules is somewhat arbitrary, and in the example above, this means the more generic rule would win out.

What now happens is the applicability rules are ordered by the length of the `Variable_Pattern`, with the assumption that the longer an applicability pattern, the more likely it is to be specific. (This isn't foolproof with complicated regular expressions, but we don't have too many of those to date). It is assumed that the more specific a regular expression is, the more likely it should take precedence. E.g., if someone write an override for a single variable, that likely should be applied instead of an override for all variables in a group (otherwise, why have the first rule at all).

I don't think this is necessarily perfect, but it feels better and more predictable behaviour than before.

varinfo/cf_config.py
```diff
@@ -160,40 +155,43 @@ def _create_attributes_object(cf_item: dict) -> dict[str, str]:
...
-        applicable attributes that apply to the specified variable. If
-        there are conflicting values for the same attribute, only the last
-        value will be returned for that attribute.
+        Last, combine the attribute names and values from each matching
+        override item. Because of the ordering in the previous step, if there
+        are multiple values supplied for the same metadata attribute, the value
+        retained will be the one with the longest variable pattern, which is a
+        proxy for how specific the override is.
 
         """
```
owenlittlejohns, line 174: Now that there are only overrides, and not supplements, there isn't a need to break out this functionality. Also the reduction in complexity of `CFConfig.get_cf_overrides` means that combining the two methods is still pretty clean.

varinfo/variable.py
```diff
@@ -242,28 +242,23 @@ def _extract_references(self, attribute_string: str) -> set[str]:
...
-        references is returned.
+        overriding dimensions from the `earthdata-varinfo` configuration
+        file, these are used instead of the raw dimensions from the `.dmr`. All
+        references are converted to absolute paths in the granule. A set of all
+        fully qualified references is returned.
 
         """
-        overrides = self.cf_config['cf_overrides'].get('dimensions')
-        supplements = self.cf_config['cf_supplements'].get('dimensions')
+        dimensions_override = self.cf_overrides.get('dimensions')
```
owenlittlejohns, line 251: When cutting out the `supplements` stuff, I thought `overrides` was a little too generic. I know the method gives context, but still the `overrides` to `dimensions_override` renaming is purely aesthetic.

varinfo/attribute_container.py
```diff
@@ -100,19 +99,13 @@ def _get_configured_attribute(
         value.
 
         """
-        cf_override = self.cf_config['cf_overrides'].get(attribute_name)
-        cf_supplement = self.cf_config['cf_supplements'].get(attribute_name)
+        cf_override = self.cf_overrides.get(attribute_name)
```
flamingbear, line 102: What if this function was just:
``` 
    return self.cf_overrides.get(attribute_name, raw_attribute_value)
```

tests/unit/test_cf_config.py
```diff
@@ -137,44 +108,45 @@ def test_get_cf_attributes_variable(self):
...
-                    config.get_cf_attributes(variable),
-                    {'cf_overrides': overrides, 'cf_supplements': supplements},
+                    config.get_cf_overrides(variable),
+                    overrides,
                 )
+
+    def test_get_cf_overrides_variable_conflicts(self):
+        """Ensure that if a variable matches multiple override rules that
+        specify conflicting values for a metadata attribute, the most specific
+        matching metadata attribute takes precedence.
```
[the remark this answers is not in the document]
flamingbear: I'm guessing I will figure this out when I understand the comments on CFConfig.get_cf_attributes

varinfo/cf_config.py
flamingbear: Yes, I was wondering if this is a good enough solution.  The thing that made the most sense to me was basically reverse order of the file. So that the "deepest" match would be used.

This would still match the "more_generic_match":
``` json
{ 
  '/this/is/my/variable': { 'attribute_name': 'more_specific_match', }, 
  '/this/is/.*': { 'attribute_name': 'more_generic_match', } 
}
```

And this would match 'more_specific_match':
``` json
{ 
  '/this/is/.*': { 'attribute_name': 'more_generic_match', }, 
  '/this/is/my/variable': { 'attribute_name': 'more_specific_match', }
  
}
```


But I expect that people will have both not put a lot of thought into how they
have structured the file and it may get pretty large, so I think sorting by
levels is a decent solution.  I kinda talked myself out of having the file
order be the important decider and I've been chewing it a few days.

This is going to trip people up no matter what.

flamingbear: > I think sorting by levels

I'm conflating two thoughts, in xarray they're sorting by depth, basically counting `/`s You're proxing specificity by length of regexp.  still seems ok.

varinfo/cf_config.py
```diff
@@ -160,40 +155,43 @@ def _create_attributes_object(cf_item: dict) -> dict[str, str]:
...
+                matching_overrides.items(),
+                key=lambda pattern: len(pattern[0]),
+            ),
+        )
 
-        return references
+        # Combine all overrides. In case of multiple overrides for the same
+        # metadata attribute, the value that comes later based on the ordering
+        # of variable patterns will be used.
+        return {
```
flamingbear, line 193:
```suggestion
        # Combine all overrides. In the case a metadata attribute appears 
        # in multiple overrides, the value that comes later based on the 
        # previous sorting of variable patterns will take precedence.
	return {
```

flamingbear (changes requested):
This looks great for the most part.  Cleans things up a lot.  I do still have a bit of confusion over when a more specific override is used vs. when multiple matches include mutiple overrides, but that's probably a me problem that you can explain in a few words.

I made a lot of suggestions where I thought the wording could be clearer, these are mostly take are leave as you feel.  I did think I simplified one of the functions, but maybe I missed something.

I'm requesting changes, but mostly it's for discussion. Looks great

config/1.0.0/earthdata_varinfo_configuration_schema.json
```diff
@@ -30,7 +30,7 @@
       }
     },
     "Excluded_Science_Variables": {
-      "description": "If CF_Overrides and CF_Supplements are not enough VarInfo classes will assume certain variables are science variables. This setting provides variable path patterns that should be excluded from consideration as a science variable.",
+      "description": "If CF_Overrides are not enough VarInfo classes will assume certain variables are science variables. This setting provides variable path patterns that should be excluded from consideration as a science variable.",
```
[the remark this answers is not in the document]
owenlittlejohns: Good question - I think this is probably from playing telephone way back when. I'll try to clarify, but I think it is something like "not enough to ensure specific variables are excluded".

varinfo/attribute_container.py
owenlittlejohns: Ha, yeah. Much simpler. Clearly I was still too caught up in the prior structure of the function.

tests/unit/test_cf_config.py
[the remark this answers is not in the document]
owenlittlejohns: I think I just missed updating this doc-string and blindly tweaked the formatting. It is definitely now incorrect. Will update.

varinfo/cf_config.py
owenlittlejohns: This is interesting. I was toying with some sort of depth element to the proxy. I guess there are two places I could see a pure length-based proxy falling down:

```
/variable_name

vs

/(variable|other_pattern).*
```

Or:

```
/group/variable
/(really|long|group|match).*
```

Doing something that first finds the patterns with the most slashes, and then finds the longest string of those patterns, would prevent the second instance, but we're still going to have problems with the first. But, hey, it would still be better. Will have a play.

config/1.0.0/earthdata_varinfo_configuration_schema.json
```diff
@@ -296,20 +289,20 @@
...
       "type": "object",
       "properties": {
         "_Description": {
-          "description": "A description of the need for and effect of the CF Convention overrides or supplements.",
+          "description": "A description of the need for and effect of the overrides.",
           "type": "string"
         },
         "Applicability": {
-          "description": "An applicability rule that can be used to define global variables, if only a mission and/or collection short name is specified, or set variable metadata attributes if a Variable_Pattern is provided.",
+          "description": "An applicability rule that can be used to define global variables, if only a mission and/or collection short name is specified, or set variable and group level metadata attributes if a Variable_Pattern is provided.",
```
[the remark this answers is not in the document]
owenlittlejohns: Looks like this description probably should have been updated when the hierarchy was updated. I _think_ that the default is that if a `Variable_Pattern` isn't given, then `.*` is used. (See [here](https://github.com/nasa/earthdata-varinfo/blob/main/varinfo/cf_config.py#L174))

I updated the wording in: [b0d29aa](https://github.com/nasa/earthdata-varinfo/pull/36/commits/b0d29aaa9e91be8a5165733060887a9902bb4099).

varinfo/cf_config.py
owenlittlejohns: Okay - this might be a bit more confusing, but I think it's a little bit better:

[edf9db6](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3)

TL;DR: Depth is now the primary measure of specificity, then string length breaks the tie within a given hierarchical depth.

tests/unit/test_cf_config.py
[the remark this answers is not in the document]
owenlittlejohns: I thought I got this in [edf9db6](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3), but didn't finish my thought. So I _really_ got it in [54b6bf1](https://github.com/nasa/earthdata-varinfo/pull/36/commits/54b6bf172569eac96fb4b8635f90c50d16f2fc9b)

owenlittlejohns (commented):
I think I caught everything. Because I changes the lines you comment on, they've gone away from the diff, but I can still go back and comment individually on the conversation threads. (Sorry that means you'll get more emails!)

varinfo/cf_config.py
```diff
@@ -160,40 +155,55 @@ def _create_attributes_object(cf_item: dict) -> dict[str, str]:
...
+        # Order override items by:
+        # First: Depth of the variable hierarchy included in the regular
+        # expression pattern (number of slashes), shallowest to deepest.
+        # Second: The length of the variable pattern, from shorted to longest.
+        # The second ordering is applied in to regular expressions specifying
+        # the same depth of variable hierarchy.
+        sorted_overrides = dict(
+            sorted(
+                matching_overrides.items(),
+                key=lambda pattern: (pattern[0].count('/'), len(pattern[0])),
```
owenlittlejohns, line 198: This is probably worth a separate comment - the `sorted` function can have a tuple returned from the key function. The documentation isn't super clear, but it seems that the ordering is done primarily on the first element of that tuple, followed by the second element. (I've beefed up the tests a bit, and things seem to be going well in those)

CHANGELOG.md
```diff
@@ -16,6 +16,22 @@ an input file.
...
+* CF Overrides retrieved for a matching file path are ordered such that the
+  most specific applicable override to the variable takes precedence. For
+  example, when requesting the value of the "units" metadata attribute for
+  variable "/nested/variable", an applicability rule that exactly matches this
+  variable path will take precedence over rules matching to either the group,
+  or all variables in the file.
+* `CFConfig.get_cf_attributes` has been renamed `CFConfig.get_cf_overrides`, as
+  there are now only overrides to be returned from this method.
+  `CFConfig.get_cf_overrides` now _must_ specify a variable path. All overrides
+  from a configuration file for a given collection is now retrievable from
```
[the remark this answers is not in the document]
owenlittlejohns: I think I got this in [edf9db](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3).

config/1.0.0/earthdata_varinfo_configuration_schema.json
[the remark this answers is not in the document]
owenlittlejohns: I heavily revamped the wording in [edf9db](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3). I _think_ it's better now, but let me know if not.

config/1.0.0/earthdata_varinfo_configuration_schema.json
```diff
@@ -239,7 +232,7 @@
           "type": "string"
         },
         "Value": {
-          "description": "The metadata attribute value. For CFOverrides, this will replace any value present, for CFSupplements, this is expected to be a string and it will be appended to the present value using string concatenation.",
+          "description": "The metadata attribute value. This will replace any value present.",
```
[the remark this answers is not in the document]
owenlittlejohns: I'm not going to comment in each thread relating to this line, but I had a crack at rewording this content in [edf9db](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3).

Definitely agree it needed something changing. Hopefully it's better now.

varinfo/attribute_container.py
owenlittlejohns: Updated in [edf9db](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3).

varinfo/cf_config.py
owenlittlejohns: Also done in [edf9db](https://github.com/nasa/earthdata-varinfo/pull/36/commits/edf9db6e153c193546a7909d76876303c689f0f3).

varinfo/cf_config.py
flamingbear: I think we are going to hope it does the right thing for most people and then point at the code when it doesn't, but this seems good to me.

varinfo/attribute_container.py
flamingbear: to be fair, I only remembered my last cleanup attempt.

flamingbear (approved):
This is excellent. Thanks for the clarifications. Looks good

PR #38, opened 2024-09-10
Trt 556 final schema clean up

owenlittlejohns:
## Description

This PR performs the final clean-up relating to the recommendations for cleaning up the `earthdata-varinfo` configuration file schema (recommendations 4 to 7 [here](https://wiki.earthdata.nasa.gov/display/SITC/earthdata-varinfo+configuration+schema+v2.0.0+proposal)). Specifically:

* `Grid_Mapping_Data` is removed (as this information wasn't directly used, and can be included as metadata attributes in `CFOverrides`.
* `ProductEpochs` is removed (as this information wasn't directly used, and can be included as a `units` metadata attribute in the more CF Convention compliant manner via `CFOverrides`).
* The schema now uses `PascalCase` for all attributes of the schema.

This is mostly small stuff updates, and I think it closes out the last of the proposed schema changes.

## Jira Issue ID

TRT-556

## Local Test Steps

* Pull this branch.
* Pip install all the local dependencies.
* Try and parse a local NetCDF-4 or DMR file. It should all work.

```
# Install dependencies:
$ pip install -r requirements.txt

# Install this version of the package (optional, if running from the root of the directory):
$ pip install -e .

# Parse a file (with/without a configuration file, this is now in a Python repl):
> from varinfo import VarInfoFromNetCDF4
>
> vi_instance = VarInfoFromNetCDF4('path/to/file.nc4', short_name='<collection short name>', config_file='/path/to/config.json')
```

## PR Acceptance Checklist
* ~~`VERSION` updated if publishing a release.~~ (This is a feature branch)

### DIFF
--- a/config/1.0.0/earthdata_varinfo_configuration_schema.json
+++ b/config/1.0.0/earthdata_varinfo_configuration_schema.json
@@ -43,11 +43,11 @@
         "$ref": "#/$defs/MissionVariablePatternType"
       }
     },
-    "CFOverrides": {
+    "MetadataOverrides": {
       "description": "# For cases where CF references do not exist, or are invalid. For example, variables that have no dimension references in the HDF-5 file contents",
       "type": "array",
       "items": {
-        "$ref": "#/$defs/CFOverridesItemType"
+        "$ref": "#/$defs/MetadataOverridesItemType"
       }
     }
   },
@@ -126,7 +126,7 @@
       "required": ["Applicability", "VariablePattern"],
       "additionalProperties": false
     },
-    "CFOverridesItemType": {
+    "MetadataOverridesItemType": {
       "description": "An item that details one or more metadata attributes to overwrite according to the supplied applicability rules.",
       "type": "object",
       "properties": {
--- a/config/1.0.0/earthdata_varinfo_configuration_schema.json
+++ b/config/1.0.0/earthdata_varinfo_configuration_schema.json
@@ -14,7 +14,7 @@
       "description": "A numeric identifier for the version of the specific configuration file (not the schema version itself).",
       "type": "integer"
     },
-    "Collection_ShortName_Path": {
+    "CollectionShortNamePath": {
       "description": "A list of HDF metadata attribute paths that provide the shortname value of the collection for the data file being processed. Processed in the listed order.",
       "type": "array",
       "items": {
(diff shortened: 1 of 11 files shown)


config/CHANGELOG.md
```diff
@@ -11,6 +11,15 @@ the earthdata-varinfo configuration file.
 This version of the configuration file schema makes several significant changes
 to simplify the schema for broader use.
 
+### Changed:
+
+* The casing of all attributes in the configuration file schema has been
+  updated to be `PascalCase` throughout. Affected schema attributes include:
+  * `CFOverrides`
```
owenlittlejohns, line 18: Something I've been toying with through the last couple of PRs... is `CFOverrides` the best name now? Part of me thinks `MetadataOverrides` would be better, but that feels like a big change. I'd love to hear some thoughts here.

tests/unit/test_var_info.py
```diff
@@ -20,9 +20,8 @@ def setUpClass(cls):
         tests.
 
         """
-        cls.config_file = 'tests/unit/data/test_config.json'
+        cls.test_config_file = 'tests/unit/data/test_config.json'
         cls.namespace = 'namespace_string'
-        cls.sample_config = 'config/0.0.1/sample_config_0.0.1.json'
```
owenlittlejohns, line 25: It felt off that some of the tests were coupled to the sample configuration file from the `config` directory. I've added the minimal amount of information to the `test_config.json` file that is used everywhere else to make all the tests pass (primarily adding a couple of short name paths, and a couple of entries in the mapping between mission and short name). Now all the tests that use a configuration file use the one in the test directory, which feels better.

config/1.0.0/earthdata_varinfo_configuration_schema.json
```diff
@@ -29,175 +29,29 @@
...
-            "$ref": "#/$defs/AngleAxisEnum"
-          },
-          "towgs84": {
-            "description": "This indicates a list of up to 7 Bursa Wolf transformation parameters., which can be used to approximate a transformation from the horizontal datum to the WGS84 datum. More precise datum transformations can be done with datum shift grids. Represented as a double-precision array, with 3, 6 or 7 values (if there are less than 7 values the remaining are considered to be zero). Corresponds to a OGC WKT TOWGS84 node.",
-            "type": "number"
-          }
-        },
-        "required": ["Grid_Mapping_Dataset_Name", "grid_mapping_name"]
-      }
-    },
```
owenlittlejohns, line 191: Looks like a big change to the schema. The biggest contribution to the lines deleted are the attributes allowed in the `Grid_Mapping_Data` section, which are a limited set of attributes, whereas the `CFOverrides` (or `MetadataOverrides`?) cover a much broader array of metadata attributes, so it becomes untenable to define all the options for attributes there.

config/1.0.0/earthdata_varinfo_configuration_schema.json
```diff
@@ -29,175 +29,29 @@
...
-    "Excluded_Science_Variables": {
-      "description": "VarInfo classes currently assume that any variable that has a grid mapping attribute, or has a spatial or temporal dimension and is not itself a dimension or bounds variable, should be treated as a science variable. This may not be true in all cases, and so Excluded_Science_Variables provide a method to denote non-science variables that might otherwise be incorrectly identified.",
+    "ExcludedScienceVariables": {
+      "description": "VarInfo classes currently assume that any variable that has a grid mapping attribute, or has a spatial or temporal dimension and is not itself a dimension or bounds variable, should be treated as a science variable. This may not be true in all cases, and so ExcludedScienceVariables provide a method to denote non-science variables that might otherwise be incorrectly identified.",
       "type": "array",
       "items": {
         "$ref": "#/$defs/MissionVariablePatternType"
       }
     },
-    "Required_Fields": {
```
owenlittlejohns, line 39: Sorry - I meant to add a comment here with the rest. The use of "Fields" here is out of sync with the rest of the schema, which uses "Variables" throughout for the same concept. I wanted to make this more consistent, particularly with `ExcludedScienceVariables`.

config/CHANGELOG.md
flamingbear: It's only in the codebase 19 times.  It is a big change, but is it better? Probably.

config/1.0.0/earthdata_varinfo_configuration_schema.json
flamingbear: Are you losing any of that information? or was most of it copied from somewhere else in the first place?  I'm ok with it all going since it can be added as a MetadataOverrides object.

config/1.0.0/sample_config_1.0.0.json
```diff
@@ -18,96 +18,151 @@
...
         "Mission": "GEDI"
       },
-      "Variable_Pattern": [
+      "VariablePattern": [
         ".*shot_number"
       ]
     }
   ],
-  "ProductEpochs": [
+  "CFOverrides": [
```
flamingbear, line 43: do it.  `MetadataOverrides`

config/1.0.0/sample_config_1.0.0.json
```diff
@@ -18,96 +18,151 @@
...
-    }
-  ],
-  "Grid_Mapping_Data": [
-    {
-      "Grid_Mapping_Dataset_Name": "EASE2_Global",
-      "grid_mapping_name": "lambert_cylindrical_equal_area",
-      "standard_parallel": 30.0,
-      "longitude_of_central_meridian": 0.0,
-      "false_easting": 0.0,
-      "false_northing": 0.0
```
flamingbear, line 64: So this is all just deleted, and not added because it wasn't used?

config/CHANGELOG.md
owenlittlejohns: I think the main client impact is really the change in the configuration file section name.

I think the only other two user-facing places that would be impacted are the `CFConfig.cf_overrides` attribute and the `CFConfig.get_cf_overrides` method (I think everything else is just plumbing). If I'm honest, my guess is that those things aren't being called by too many end users.

Given that this is going to be the most disruptive change to the package, now is the time to do it. I think I'm leaning to making this change. (And if it seems terrible, I can always revert the commit before merging this PR)

flamingbear (approved):
Looks great, one question about dropping stuff from the sample config (but I'm pretty sure the answer is yet) and I'd rename CFOverrides now because just because we use it for that, it's not specific to CF.

config/1.0.0/earthdata_varinfo_configuration_schema.json
owenlittlejohns: All the specific attributes definitions were just copied from the CF Conventions page itself.

config/1.0.0/sample_config_1.0.0.json
owenlittlejohns: The `ProductEpoch` was entirely unused, so I just deleted it. I moved the content from `Grid_Mapping_Data` to the `CFOverrides`/`MetadataOverrides` section, though, just to provide examples for how that information would be included in the future. (In theory that sample configuration file should probably be slimmed down to a minimum number of examples, but I think that some of those rules were placed there so DAS has them somewhere)

config/1.0.0/sample_config_1.0.0.json
owenlittlejohns: [578faf8](578faf8974eb086f0f60841a47d70f1ae47a008e) :tada:

owenlittlejohns (commented):
Quick summary of new changes:

* Rename `CFOverrides` to `MetadataOverrides`.
* Rename `CFConfig.cf_overrides` to `CFConfig.metadata_overrides`.
* Rename `CFConfig.get_cf_overrides` to `CFConfig.get_metadata_overrides`.
* Rename the `cf_overrides` property on all the `Attribute`/`Group`/`Variable` classes to be `metadata_overrides`. (That sounds like a lot, but it's mostly sorted via inheritance)
* Review change logs (both schema and overall package) and reword/reorder things based on importance rather than the ordering in which they've been done on this feature branch.

eni-awowale (approved):
Tested with `ATL06` and worked as expected!

PR #41, opened 2024-10-21
MSD-4058 - Replace native ID  spaces with underscores.

owenlittlejohns:
## Description

This PR supports ASDC, who are trying to use `earthdata-varinfo` to generate UMM-Var records for variables that include spaces in their full paths. This is currently causing issues in the choice of CMR native ID. To support this, spaces in generated native IDs are converted to underscores (in a similar manner to how slashes are already converted to underscores).

As a bonus, this PR relaxes the `numpy` dependency, which theoretically allows for support under Python 3.12.

~~This cannot be rigorously tested via our CI/CD at this time, because one of the dependencies our testing uses (`xmlrunner`) does not work under Python 3.12. However, continued operations under Python 3.9, 3.10 and 3.11 can be confirmed.~~ 

**Update:** I found you can run a test suite that utilises `unittest` style classes with `pytest`, and so use the plugins from the `pytest` ecosystem. I've swapped over to running our `unittest` style test suite with `pytest`, which allows us to use the `--junitxml` option. This does not use `unittest-xml-runner` and is compatible with Python 3.12 :tada:. This PR also now uses `pytest-cov` to generate coverage reports as before.

Generally, providing a broader range of options for `numpy` enables `earthdata-varinfo` to be compatible with more client Python environments. The relaxed constraint continues to support the previously pinned version at a minimum.

## Jira Issue ID

MSD-4058 (not part of the regular ESDIS/EED Jira instance)

## Local Test Steps

* Pull this branch.
* Use `earthdata-varinfo` as you would have before, to make sure there are no regressions.
* Try generating some UMM-Var records for a file with spaces in the variables names. Then use the `generate_variable_native_id` function to confirm that any spaces are converted to underscores.

## PR Acceptance Checklist
* ~~Documentation updated (if needed).~~


owenlittlejohns:
@hazemmahmoud88 - here is the change we discussed in Slack. (I don't know Ryan's GitHub user to tag him, too)

varinfo/umm_var.py
```diff
@@ -403,7 +403,10 @@ def generate_variable_native_id(
 
     """
     return '-'.join(
-        [collection_concept_id, umm_var_record['Name'].replace('/', '_').lstrip('_')]
+        [
+            collection_concept_id,
+            umm_var_record['Name'].replace('/', '_').replace(' ', '_').lstrip('_'),
```
owenlittlejohns, line 408: This is the primary change for the PR.

varinfo/cmr_search.py
```diff
@@ -36,7 +37,7 @@ def get_granules(
     provider: str | None = None,
     cmr_env: CmrEnvType = CMR_OPS,
     auth_header: str | None = None,
-) -> list:
+) -> Sequence:
```
owenlittlejohns, line 40: This is a sneaky extra change I'm getting in here due to updating `python-cmr` to a version with type hints.

I'm not a big fan of using `Sequence` here, because I think it really is a `list`, and that's more specific, but the underlying type hint for the return type from `cmr.GranuleQuery.get` is a `Sequence`, so I changed our type hint to be compatible with that.

(The other alternative was to explicitly cast the output of `cmr.GranuleQuery.get` as a `list` - I could be convinced to go either way)

eni-awowale:
Hey Owen! Let me know if I am missing something. Is this PR supposed to update the UMMVar record to not have spaces for the values in the field `"Name"`?

owenlittlejohns:
@eni-awowale - This PR is only going to affect the native ID that is used with CMR to not include spaces if the `Name` attribute of the generated UMM-Var record does. That `Name` should remain unchanged. Here's an example screenshot that Ryan Omidi (ASDC) provided, with examples of spaces on the right.

image

In this example we'd end up with an example variable name of `Time and Position/Packet number`. That will still be the case both before and after this PR. The Native ID that CMR uses would have been: `C1234567890-PROV-Time and Position_Packet number`, but the CMR ingest API doesn't allow for those spaces in the native ID. Instead, this PR will lead to a native ID of `C1234567890-PROV-Time_and_Position_Packet_number`.

The main reason for this change is having spaces in the native IDs was causing errors when making calls to the CMR ingest API.

For what it's worth, I was a little surprised that there were files out there that included spaces in group names and/or variable names. But hey, now we should be able to ingest these sorts of things.

eni-awowale:
Thanks for clarifying! There are a few GES_DISC collections that have spaces in variable names: OMSO2_003, OMNO2_003 etc. I tested this with OMNO2_003 and it returned `'C1239966837-GES_DISC-HDFEOS_SWATHS_ColumnAmountNO2_Data_Fields_AMFQualityFlags'`.

eni-awowale (approved):

PR #51, opened 2025-02-10
Das 2287 earthdata varinfo var dim shape

vutrannasa:
## Description

Support for `VariableFromDmr` variable dimension shape data in NetCDF-4 files with named dimensions and HDF-5 files with anonymous size-only dimensions.

Changes:
- `VariableFromDmr::_get_shape()` returns dimension shape data for NetCDF-4
files with named dimensions and HDF-5 files with anonymous size-only dimensions.
- Update DRM `unittest` to validate variable dimension shape data.

## Jira Issue ID

[DAS-2287](https://bugs.earthdata.nasa.gov/browse/DAS-2287)

## Local Test Steps
Local MAC M1

Verify unit test
```
% cd /<your_workspace>
% git clone https://github.com/nasa/earthdata-varinfo.git DAS-2287-earthdata-varinfo-var-dim-shape
% cd DAS-2287-earthdata-varinfo-var-dim-shape
% conda create --name earthdata-varinfo python=3.11 --channel conda-forge --channel defaults -y
% conda activate earthdata-varinfo
% pip install -r requirements.txt -r dev-requirements.txt
% make test
```

Test notebook
 - Load VScode and open DAS-2287-earthdata-varinfo-var-dim-shape folder
 -- Select doc/earthdata-varinfo.ipynb
 -- Click "Run All" button
 -- Verify test runs without issues

## PR Acceptance Checklist
* [ x] Jira ticket acceptance criteria met.
* [ x] `CHANGELOG.md` updated to include high level summary of PR changes.
* [ x] `VERSION` updated if publishing a release.
* [ x] Tests added/updated and passing.
* [ x] Documentation updated (if needed).


config/1.0.0/sample_config_1.0.0.json
```diff
@@ -292,7 +292,7 @@
     {
       "Applicability": {
         "Mission": "ICESat2",
-        "ShortNamePath": "ATL0[3-9]|ATL1[023]",
+        "ShortNamePath": "ATL03",
```
vutrannasa, line 295: This was a fix found by David in DAS-2289
"/gt[123][lr]/geolocation/." is only ATL03

https://git.earthdata.nasa.gov/projects/SITC/repos/trajectorysubsetter/pull-requests/84/overview?commentId=98409

docs/earthdata-varinfo.ipynb
```diff
@@ -106,10 +107,11 @@
    "source": [
     "import json\n",
     "\n",
-    "from varinfo import VarInfoFromNetCDF4\n",
+    "from varinfo import VarInfoFromNetCDF4, VarInfoFromDmr\n",
```
vutrannasa, line 110: The notebook tests VarInfoFromNetCDF4 and VarInfoFromDmr using the same granule with two different input files (HDF5, DMR) to ensure consistent results. 

gpm_granule_path = B-HHR.MS.MRG.3IMERG.20200201-S233000-E235959.1410.V06B.HDF5 gpm_dmr_granule_path = GPM_3IMERGHH.06_3B-HHR.MS.MRG.3IMERG.20200201-S233000-E235959.1410.V06B.HDF5.dmr.xml

docs/earthdata-varinfo.ipynb
```diff
@@ -434,7 +565,6 @@
     "\n",
     "The list below refers to potential improvements that can be made within the core Python package to improve the schema coverage of the generated UMM-Var records:\n",
     "\n",
-    "* Ensuring variables parsed from a DMR file also contain shape information, similar to current functionality of NetCDF-4 file parsing. This information is stored within a DMR in separate `<Dimension />` XML elements. This would allow UMM-Var JSON generated from DMR data to have sizes on their dimensions.\n",
```
vutrannasa, line 437: Remove the comment from the "earthdata-varinfo UMM-Var future work:" section, as the requirements have been fulfilled.

tests/unit/test_umm_var.py
```diff
@@ -391,17 +391,30 @@ def test_get_dimension_information(self):
...
+            # Uses M2I3NPASM from DMR
             merra_variable = self.merra_varinfo.get_variable('/EPV')
+            self.assertDictEqual(
+                get_dimension_information(self.merra_varinfo, merra_variable, '/time'),
+                {'Name': 'time', 'Size': 8, 'Type': 'TIME_DIMENSION'},
+            )
             self.assertDictEqual(
                 get_dimension_information(self.merra_varinfo, merra_variable, '/lev'),
-                {'Name': 'lev', 'Size': 'Varies', 'Type': 'OTHER'},
+                {'Name': 'lev', 'Size': 42, 'Type': 'OTHER'},
```
vutrannasa, line 406: Updated tests now include size information for VarInfoFromDmr and VariablesFromDmr

tests/unit/test_umm_var.py
```diff
@@ -411,7 +424,7 @@ def test_get_dimension_information(self):
                 ),
                 {
                     'Name': 'gt3r/bckgrd_atlas/delta_time',
-                    'Size': 'Varies',
+                    'Size': 83188,
```
vutrannasa, line 427: Updated tests now include size information for VarInfoFromDmr and VariablesFromDmr
https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/ATL03_example.dmr#L21220

tests/unit/test_umm_var.py
```diff
@@ -428,21 +441,6 @@ def test_get_dimension_information(self):
                 {'Name': 'Grid/latv', 'Size': 2, 'Type': 'OTHER'},
             )
 
-        with self.subTest('Absent dimension'):
```
vutrannasa, line 431: Remove test as the requirements have been fulfilled.

tests/unit/test_umm_var.py
```diff
@@ -496,17 +494,15 @@ def test_get_dimension_size(self):
...
+        with self.subTest('Variable with shape returns correct size'):
             # This test currently uses input from a DMR file, as the
             # VariableFromDMR class does not derive shape information for any
             # variable - this guarantees skipping the first two conditions in
             # the function.
             gpm_variable = self.gpm_varinfo.get_variable('/Grid/precipitationCal')
             self.assertEqual(
                 get_dimension_size(self.gpm_varinfo, gpm_variable, '/Grid/lon'),
-                'Varies',
+                3600,
```
vutrannasa, line 505: Updated tests now include size information for VarInfoFromDmr and VariablesFromDmr
https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/GPM_3IMERGHH_example.dmr#L27

tests/unit/test_variable.py
```diff
@@ -686,6 +682,58 @@ def test_get_valid_max(self):
 
                 self.assertEqual(variable.get_valid_max(), expected_valid_max)
 
+    def test_variable_get_shape(self):
```
vutrannasa, line 685: Add additional tests to verify VariableFromDmr variable dimension shape data.

tests/unit/test_variable.py
```diff
@@ -177,11 +177,7 @@ def test_variable_instantiation(self):
             {'/group/begin', '/group/count'},
         )
 
-        # Currently the Dimension elements from the DMR, which contain the
-        # number of elements for a variable in a given dimension, are not
-        # preserved. This means the variable shape is not tracked for DMR
-        # variables.
-        self.assertIsNone(variable.shape)
+        self.assertEqual(variable.shape, [1800, 3600])
```
vutrannasa, line 180: Remove comments and verify variable dimension shape data.

varinfo/var_info.py
```diff
@@ -565,13 +565,14 @@ def _extract_variables(self):
 
         """
 
-        def save_variable(output, group_path, element):
+        def save_variable(output, group_path, element, dim_name_size):
```
vutrannasa, line 568: Created a `dim_name_size dictionary` to store dimension names and sizes from <Dimension /> elements, allowing <Dim /> elements to reference dimension names and retrieve their corresponding sizes.
[Example](https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/dimension_grouping.dmr#L6)

From Owen's comments
Genuine NetCDF-4 files with named dimensions seem to not include the dimension size in the <Dim /> elements within a variable. They point back to <Dimension /> elements that have the size in them. I think that the Variable object doesn't currently have access to the full DMR at the time it's deriving the shape, so that was why I didn't implement anything.

varinfo/var_info.py
```diff
@@ -624,14 +625,21 @@ def traverse_elements(
...
             # function again
             element_type = child.tag.replace(self.namespace, '')
 
             if element_type in element_types:
-                operation(output, group_path, child)
+                operation(output, group_path, child, dim_name_size)
+            elif element_type == 'Dimension':
+                if child.attrib['size'] is not None:
+                    dim_name = '/'.join([group_path, child.get('name')])
+                    dim_name_size[dim_name] = int(child.attrib['size'])
```
vutrannasa, line 642: Add `dim_name_size dictionary` to store dimension names and sizes from elements.  Example of output

https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/dimension_grouping.dmr#L6
`dim_name_size dictionary` = {'/time': '1', '/latitude': '1800', '/longitude': '3600'}

varinfo/variable.py
```diff
@@ -50,9 +50,9 @@ def __init__(
         super().__init__(variable, cf_config, namespace, full_name_path)
         self.group_path, self.name = self._extract_group_and_name()
         self.data_type = self._get_data_type(variable)
-        self.shape = self._get_shape(variable)
         self.references = self._get_all_cf_references()
         self.dimensions = self._extract_dimensions(variable)
+        self.shape = self._get_shape(variable)
```
vutrannasa, line 55: When VariableFromDmr is instantiated, the `dim_name_size dictionary` will already contain the parsed dimension names and sizes.
Example: `dim_name_size dictionary` = {'/time': '1', '/latitude': '1800', '/longitude': '3600'}

First, we extract the dimension names from [Dim elements](https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/dimension_grouping.dmr#L10), then use these names to locate and retrieve the shape/size from [Dimension elements](https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/dimension_grouping.dmr#L6).

varinfo/variable.py
```diff
@@ -341,17 +341,42 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
+        """Extract the shape of the variable data array.  If there are
+        <DIM size=xxx/> elements with a size attribute, use them as the
+        variable shape instead of <Dimensions size=xxx/> from the
+        parent element. Otherwise, use <Dimensions size=xxx/> to define
+        the variable shape.
         """
 
+        # Retrieve the shape from the dimension size in the
+        # <Dim size=xx/> element.
+        self.shape = [
```
vutrannasa, line 364: To support HDF-5 files with anonymous/size-only dimensions list the dimension size directly in the <Dim /> element
Example `<Dim name="latv" size="5555"/>`

varinfo/variable.py
```diff
@@ -341,17 +341,42 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
+            int(dim_size.attrib['size'])
+            for dim_size in variable.findall(f'.//{self.namespace}Dim[@size]')
+        ]
+
+        # Retrieve the shape from dimension size in the
+        # <Dimension size=xx/> element
+        if not self.shape:
```
vutrannasa, line 371: If anonymous/size-only dimensions are not set on line 364, then assign `self.shape` to the <Dimension size=xx/> element.

https://github.com/nasa/earthdata-varinfo/blob/main/tests/unit/data/dimension_grouping.dmr#L11
From above link the output for `self.dimensions` and `self.shape` 

`self.dimensions` = ['/latitude']
`self.shape` = 1800

varinfo/variable.py
```diff
@@ -341,17 +341,47 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
-        """Extract the shape of the variable data array. This is not yet
-        implemented as the Dimension information is currently unavailable
-        to the Variable XML content.
-
+        """Extract the shape of the variable data array.  If there are
+        <DIM size=xxx/> elements with a size attribute, use them as the
+        variable shape instead of <Dimensions size=xxx/> from the
+        parent element. Otherwise, use <Dimensions size=xxx/> to define
+        the variable shape.
         """
```
D-Auty, line 365: """  Extract the shape of the variable data array.  First explore
        XML for <DIM size=xxx/> - DIM child elements with a size
        attribute - true for HDF5 files and anonymous dimensions
        (note no name). If found (with or without names), use them. If
        not found, use dim_name_size dictionary with matching dimension
        names to define the variable shape. Note the dim_name_size
        dictionary is filled when parsing xml for variables, capturing
        dimensions names and sizes for dimension entries - as found in
        NetCDF files with named dimensions.
"""

varinfo/variable.py
```diff
@@ -341,17 +341,47 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
 
     """
 
+    def __init__(
+        self,
+        element,
+        cf_config,
+        namespace,
+        full_name_path,
+        dim_name_size=[str, int],
```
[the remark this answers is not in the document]
vutrannasa: Setting `dim_name_size=[str, int]`  as an empty list in method signatures ensures compatibility with existing unit tests while also enabling support for passing NetCDF-4 dimension names and sizes to `VariableInfoFromDmr`.

https://github.com/nasa/earthdata-varinfo/pull/51/files#diff-5d585e51a7aba6de289945217bc116a605ce784ada6352fa97fdc008f26287f6R138

varinfo/variable.py
vutrannasa: Will update

varinfo/variable.py
```diff
@@ -341,17 +341,51 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
     def _get_data_type(self, variable: ET.Element) -> str:
         """Extract a string representation of the variable data type."""
         return variable.tag.lstrip(self.namespace).lower()
 
     def _get_shape(self, variable: ET.Element) -> tuple[int]:
-        """Extract the shape of the variable data array. This is not yet
-        implemented as the Dimension information is currently unavailable
-        to the Variable XML content.
-
+        """Extract the shape of the variable data array. First explore
```
owenlittlejohns, line 360: A similar/duplicate comment might show up - I wrote it on a line that got updated and then the update removed my comment.

I think this methodology could potentially be simplified.

Every variable is being provided with the mapping of `<Dimension />` names to sizes. And I think that for every `<Dim />` element in a DMR file, there is a corresponding `<Dimension />` element. I also think that the `<Dimension />` element always has a `size` attribute, whether or not the `<Dim />` does. So an alternative here would be to just always get the size from the `<Dimension />` element, making this method more like:

```
def _get_shape(self, variable: ET.Element) -> tuple[int]:
    """<doc string describing things>"""
    return [
        self.dim_name_size[dimension.attrib['name']]
        for dimension in variable.findall(f'.//{self.namespace}Dim')
    ]
```

An advantage here is that this also covers a potential edge case of a variable that has some `<Dim />` elements with a `size` attribute and others that do not. Currently, this method would return a tuple with the wrong number of elements (because `self.shape` would not be an empty list, so it would skip the second list comprehension).

varinfo/variable.py
```diff
@@ -341,17 +341,51 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
+        # Retrieve the shape from dimension size in the
+        # <Dimension size=xx/> element
+        if not self.shape:
+            self.shape = [
+                self.dim_name_size[dim_name]
+                for dim_name in self.dimensions
+                if dim_name in self.dim_name_size
+            ]
+
+        return self.shape
```
owenlittlejohns, line 387: General note (that might be OBE depending on if you update this method to always grab sizes from `<Dimension />` elements):

This method is returning the value for `shape` that is assigned to `self.shape` in the `VariableBase.__init__`. Generally, you don't want to both assign to a class attribute within a method, and then return the value to be assigned elsewhere. In this example, it would be better to make a locally scoped variable (maybe `shape`) and then let the assignment happen in `VariableBase.__init__`.

varinfo/variable.py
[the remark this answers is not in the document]
owenlittlejohns: I agree with David here - this feels a bit wonky. Isn't the input a dictionary mapping strings to integers, rather than a list?

If other tests are breaking, that makes sense - the function signature has changed. The better thing to do would probably be to update those tests.

For what it's worth, I definitely like adding type hints as we go. In this case, the type hint should probably be:

```
dim_name_size: dict[str, int]
```

varinfo/umm_var.py
```diff
@@ -226,9 +226,6 @@ def get_dimension_information(
     * Projected horizontal spatial dimension (no UMM-Var type for this).
     * Swath dimensions (along or across track).
 
-    Also note: VariableFromDmr instances do not currently extract variable
-    shapes.
-
```
owenlittlejohns, line 231: Nice - I appreciate you keeping this documentation up-to-date, particularly as it's not really in the main area of the code change.

owenlittlejohns (changes requested):
First up - I like the idea of grabbing all the `<Dimension />` elements and making a mapping containing their sizes. That seems like a lightweight way of giving information to the `VariableFromDmr` instances (rather than a full XML tree).

A couple of things:

* I think the `VariableFromDmr._get_shape` method could be simplified if we always just get the size from the `<Dimension />` elements, rather than first checking for a `size` attribute in the `<Dim />` element. I think the method now has enough information to always do this.
* I'm a bit worried about the `dim_name_size` mapping being populated at the same time as the variables are being parsed. I don't 100% trust that a DMR will always contain the `<Dimension />` elements in a place that will be parsed prior to a variable that needs them.
* I agree with David regarding the default value versus function signature. It seems like the tests should really be updated instead of adding a bit of an odd default value for the new parameter.

I think those are the big things. There's a few smaller code things, too, but those are quick things to work on.

tests/unit/test_umm_var.py
owenlittlejohns: While this change was geared towards supporting HOSS, this is a nice bonus with the UMM-Var generation!

varinfo/variable.py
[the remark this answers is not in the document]
D-Auty: Perhaps: `dim_name_size: dict[str, int] = {}`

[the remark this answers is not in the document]
owenlittlejohns: I think that `{}` (an empty dictionary) would be the correct choice for a default argument if one was to be added. But I think we should avoid adding a default argument to get tests to pass. Really, the tests should be updated to reflect the code, not the other way around.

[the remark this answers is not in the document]
D-Auty: This is relevant in other cases where dim_name_size is being added to the methods or structure. See var_info.py and _extract_variables.  While this code was not developed to the coding standard applying type hints, I would like to see these, and adding them in incrementally is an acceptable compromise to taking the time to retrofit, vs not having them at all.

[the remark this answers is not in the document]
owenlittlejohns: Totally agree we should have type hints, and totally agree adding them as we go is a good approach. (I've been eying up adding some type hints in some spare time that never comes around)

Perhaps the best approach is to add all the type hints to a method or function any time it is updated. In this case that would mean:

```
def __init__(
        element: ET.Element,
        cf_config: CFConfig,
        namespace: str,
        full_name_path: str,
        dim_name_size: dict[str, int],
):
```

That is a distinct conversation from adding a default value, which has a function impact on the code. (I think we're on the same page here, but I want to reiterate it, to be sure)

varinfo/var_info.py
```diff
@@ -624,14 +625,21 @@ def traverse_elements(
 
         group_path = group_path.rstrip('/')
 
+        dim_name_size = {}
```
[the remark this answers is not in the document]
vutrannasa: Change to `all_dimensions_sizes `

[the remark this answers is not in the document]
vutrannasa: It's possible, currently `all_dimensions_sizes` local since it acts as a stack. When `element_type == 'Group'`, it invokes the recursive method [traverse_elements()](https://github.com/nasa/earthdata-varinfo/blob/DAS-2287-earthdata-varinfo-var-dim-shape/varinfo/var_info.py#L646
) and creates a new instance of `all_dimensions_sizes` for the Group element, as it may contain its own <Dimension>

[ATL03_example.dmr](https://github.com/nasa/earthdata-varinfo/blob/DAS-2287-earthdata-varinfo-var-dim-shape/tests/unit/data/ATL03_example.dmr#L3)
[Group](https://github.com/nasa/earthdata-varinfo/blob/DAS-2287-earthdata-varinfo-var-dim-shape/tests/unit/data/ATL03_example.dmr#L2214C1-L2215C56)
```
<Dataset xmlns="http://xml.opendap.org/ns/DAP/4.0#" xml:base="http://opendap.uat.earthdata.nasa.gov/providers/EEDTEST/collections/ATLAS-ICESat-2%20L2A%20Global%20Geolocated%20Photon%20Data%20V003/granules/EEDTEST-ATL03-003-ATL03_20200714T235814" dapVersion="4.0" dmrVersion="1.0" name="ATL03_20200714235814_03000802_003_01.h5">
    <Dimension name="ds_surf_type" size="5"/>
    <Dimension name="ds_xyz" size="3"/>
...
        <Group name="bckgrd_atlas">
            <Dimension name="delta_time" size="83140"/>
```

varinfo/variable.py
```diff
@@ -341,17 +341,53 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
+        if (
+            len(shape) != len(self.all_dimensions_sizes)
+            and len(self.all_dimensions_sizes) > 0
+        ):
+            for dim_name in self.dimensions:
+                if dim_name in self.all_dimensions_sizes:
+                    index = self.dimensions.index(dim_name)
+                    shape.insert(index, self.all_dimensions_sizes[dim_name])
+
+        return shape
```
vutrannasa, line 389: Added code to support ordered shapes for anonymous/size-only dimensions in the <Dim /> element. 
**Example:**
```
  <Dimension name="latitude" size="1111"/>
  <Dimension name="longitude" size="2222"/>
  <Float64 name="lat_bnds">
    <Dim name="latitude"/>
    <Dim name="latv" size="5555"/>
   ...
  </Float64>
  <Float64 name="lon_bnds">
    <Dim name="lonv" size="7777"/>
    <Dim name="longitude"/>
   ...
  </Float64>
```

For the variable "lat_bnds," the <Dim /> elements include latitude (without a size, retrieved from parent Dimension) and latv (size = 5555). The code correctly outputs the shape order as [1111, 5555].
```
Variable: lat_bnds
self.dimensions: ['/science_three/latitude', '/science_three/latv']
self.all_dimensions_sizes: {'/science_three/latitude': 1111, '/science_three/longitude': 2222}
shape: [1111, 5555]
```

For the variable "lon_bnds," the <Dim /> elements include lonv (size = 7777) and latitude (without a size, retrieved from parent Dimension) . The code correctly outputs the shape order as [7777, 2222].
```
Variable: lon_bnds
self.dimensions: ['/science_three/lonv', '/science_three/longitude']
self.all_dimensions_sizes: {'/science_three/latitude': 1111, '/science_three/longitude': 2222}
shape: [7777, 2222]
```

varinfo/var_info.py
```diff
@@ -624,14 +625,21 @@ def traverse_elements(
...
             # If it is in the DAP4 list: use the function
+            # else, if it is a Dimension, assign dictionary dim_name_size{}
             # else, if it is a Group, assign to dictionary and call this
             # function again
             element_type = child.tag.replace(self.namespace, '')
 
             if element_type in element_types:
-                operation(output, group_path, child)
+                operation(output, group_path, child, dim_name_size)
+            elif element_type == 'Dimension':
```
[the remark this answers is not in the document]
owenlittlejohns: @vutrannasa - I think this comment remains a question. Did you and @D-Auty confirm that a `<Dimension />` element in a DMR file _has_ to precede any variable that makes reference to it?

docs/earthdata-varinfo.ipynb
```diff
@@ -128,7 +130,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "gpm_imerg = VarInfoFromNetCDF4(gpm_granule_path, short_name='GPM_3IMERGHH')"
+    "# Instantiate VarInfoFromNetCDF4\n",
+    "gpm_imerg = VarInfoFromNetCDF4(gpm_granule_path, short_name='GPM_3IMERGHH')\n",
+    "\n",
+    "# Instantiate VarInfoFromDmr\n",
+    "gpm_imerg_dmr = VarInfoFromDmr(gpm_dmr_granule_path, short_name='GPM_3IMERGHH')"
```
owenlittlejohns, line 137: It's definitely a good addition to show a user they can use `VarInfoFromDmr` or `VarInfoFromNetCDF4` in equivalent ways. Also - thanks for paying attention to the documentation notebook. (It's so easy for this sort of thing to get stale!)

That said, would the easiest thing be to just have this one cell with something like:

```
"# Instantiate VarInfoFromNetCDF4\n",
    "gpm_imerg = VarInfoFromNetCDF4(gpm_granule_path, short_name='GPM_3IMERGHH')\n",
    "\n",
    "# Instantiate VarInfoFromDmr\n",
    "# gpm_imerg = VarInfoFromDmr(gpm_dmr_granule_path, short_name='GPM_3IMERGHH')"
```

Then in the instructions above, tell the user they can do the same things using either class, and to un-comment the appropriate line declaring `gpm_imerg` using the class they want to use. After that you'll only need the one set of cells as they were before.

I'm mainly thinking that we don't need a side-by-side comparison, as much as just a way to let a user run through the example cells using either `VarInfoFromNetCDF4` or `VarInfoFromDmr`.

(Feel free to disagree, I'm just wondering how to keep the documentation as simple as possible)

owenlittlejohns (commented):
This is looking pretty close. The main outstanding things:

1) The question about whether a `<Dimension />` element is guaranteed to be earlier in a DMR than any variable that refers to it.
2) Adding a quick subtest for the mixed with/without size case.
3) (Something I don't have too strong opinions on) Possibly removing some of the duplicate cells in the documentation notebook.

Everything else looks good. Thanks for addressing a bunch of my comments, and thanks in particular for going back and updating all the tests updated by the change in function signature for `VariableFromDmr`.

varinfo/variable.py
owenlittlejohns: Marking as resolved. Thanks for addressing this.

varinfo/var_info.py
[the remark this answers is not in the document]
owenlittlejohns: Oh... wait, that's a good catch:

> it invokes the recursive method [traverse_elements()](https://github.com/nasa/earthdata-varinfo/blob/DAS-2287-earthdata-varinfo-var-dim-shape/varinfo/var_info.py#L646) and creates a new instance of all_dimensions_sizes for the Group element

I think this is something we _don't_ want to do. I think we want one global dictionary of dimensions as child groups can (and do) refer to dimensions in their parents. Here's [an example DMR](https://opendap.uat.earthdata.nasa.gov/collections/C1234714698-EEDTEST/granules/SC:ATL08.005:230405690.dmr.xml) from ICESat-2/ATL08. The `/ds_metrics` dimension in the root group is used by `/gt1l/land_segments/canopy/canopy_h_metrics` and `/gt1l/land_segments/canopy/canopy_h_metrics_abs` (and similar variables in the other ground track groups).

We should also have a test for this, where something in `dimension_grouping_size.dmr` is a nested variable referring to a dimension outside of its own immediate group (so maybe a new variable in the group `/science_four` that refers to a dimension in the root group of the `Dataset`).

owenlittlejohns (changes requested):
Sorry for the quick-fire second review. I was resolving my comments that you'd taken care of and saw what you put [here](https://github.com/nasa/earthdata-varinfo/pull/51#discussion_r1953597542).

I think this is definitely something that needs to be tweaked. The `all_dimensions_sizes` dictionary needs to contain all dimensions from within a granule, not just those in the single group. (See ATL08 example linked in my reply in that thread)

varinfo/var_info.py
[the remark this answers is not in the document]
vutrannasa: @D-Auty and I had a conversion after we talked.  I believe your statement is correct, David can you confirm again?

varinfo/var_info.py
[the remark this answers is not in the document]
vutrannasa: Lets use [DAS-2305](https://bugs.earthdata.nasa.gov/browse/DAS-2305) in the next sprint 25.3 to address this fix.

docs/earthdata-varinfo.ipynb
vutrannasa: As a new user or developer, I would like to see how two different inputs (NetCDF and DMR) produce the same output side by side. While combining them saves space, being able to verify both outputs on the same screen is more convenient than capturing the NetCDF output separately and then rerunning the process for the DMR input.

tests/unit/test_variable.py
```diff
@@ -681,14 +757,73 @@ def test_get_valid_max(self):
...
 
+    def test_variable_get_shape(self):
+        """Ensure that all variable shapes are returned when requesting
+        variable.shape and dimension.
+
+        """
+        with self.subTest('<Dim /> element with name and size variable'):
+            dmr_variable = ET.fromstring(
+                f'<{self.namespace}Float64 name="science">'
+                f' <{self.namespace}Dim name="time" size="2"/>'
```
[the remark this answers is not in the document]
vutrannasa: Add two extra tests to hand Dim without name
- https://github.com/nasa/earthdata-varinfo/blob/DAS-2287-earthdata-varinfo-var-dim-shape/tests/unit/test_variable.py#L811

- https://github.com/nasa/earthdata-varinfo/blob/DAS-2287-earthdata-varinfo-var-dim-shape/tests/unit/test_variable.py#L846

varinfo/var_info.py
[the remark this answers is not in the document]
D-Auty: I can absolutely confirm that the group_one  / group_two use-case suggested above is not valid.  NetCDF requires dimensions to be strictly in the hierarchy - no "up and over" as we sometimes see with coordinates.  This is because in the NetCDF creation (and storage I suspect), the dimension name is not captured with a group reference, only as a "bare" dimension name.  The resolution can only be in the same group or as an "up-level" reference.  I believe it is possible for a group dimension to override an up-level reference, but I'm not sure even that is allowed, and I don't think it would be considered good practice.

As for DMR xml capture of the NetCDF data architecture, I'm pretty sure, but cannot be as certain we will never encounter the case suggested above - where a dimension reference precedes the dimension definition.  We have no examples of such a case, and it seems unlikely it would be created this way, but that is not certainty.

docs/earthdata-varinfo.ipynb
owenlittlejohns: > As a new user or developer

I think this is the crux of our difference of opinion. This documentation is primarily geared towards an end user, not a developer. I think the comparison of the two classes is primarily of benefit to a developer, and not as important to an end user; an end-user just wants to know what one of the classes can do, not that the output from two classes looks the same.

That said, I don't think it's massively important. I'll mark this thread as resolved once the review is published.

varinfo/var_info.py
```diff
@@ -711,6 +747,9 @@ def _parse_group(self, group: Dataset | Group):
                 full_name_path=variable_path,
             )
 
+            for index, dimensions_name in enumerate(variable.dimensions):
+                self.all_dimensions_sizes[dimensions_name] = variable.shape[index]
```
owenlittlejohns, line 751: `enumerate` is a good way to populate `VarInfoFromNetCDF4.all_dimension_sizes`, but this class doesn't have this attribute as you declare it in `VarInfoFromDmr.__init__` and then use `super` to add in all the basic stuff from `VarInfoBase`. `all_dimension_sizes` is also not used anywhere in the rest of the `VarInfoFromNetCDF4` or `VariableFromNetCDF4` classes, so I don't think you need these lines (unless I'm missing something).

There are two possible resolutions here:

1) Remove these lines.
2) Add `all_dimension_sizes` to `VarInfoBase.__init__` rather than in `VarInfoFromDmr.__init__`.

varinfo/variable.py
```diff
@@ -341,17 +341,53 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
...
+        shape = [
+            int(dim_size.attrib['size'])
+            for dim_size in variable.findall(f'.//{self.namespace}Dim[@size]')
+        ]
+
+        # Retrieve the shape from dimension size in the
+        # <Dimension size=xx/> element
+        if (
+            len(shape) != len(self.all_dimensions_sizes)
+            and len(self.all_dimensions_sizes) > 0
```
owenlittlejohns, line 382: This seems off:

`self.all_dimension_sizes` is a cache of all dimensions for the DMR, not a measure of the total number of dimensions of the variable being evaluated. What we're trying to check here is: does the `shape` list have an element for every dimension of the variable array.

The check should be updated to check whether the length of `shape`, which is the number of `<Dim /> elements with a `size` attribute, is the same as the total number of `<Dim />` elements for the variable.

varinfo/variable.py
```diff
@@ -341,17 +341,42 @@ class VariableFromDmr(VariableBase, AttributeContainerFromDmr):
 
     """
 
+    def __init__(
+        self, element, cf_config, namespace, full_name_path, dim_name_size=[str, int]
+    ):
+        self.dim_name_size = dim_name_size
+        super().__init__(element, cf_config, namespace, full_name_path)
```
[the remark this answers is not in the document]
owenlittlejohns: This feels like the right place to declare `all_dimension_sizes` now. Thanks! (I'll resolve the relevant conversation thread after publishing my review)

varinfo/var_info.py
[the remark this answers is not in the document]
owenlittlejohns: Okay - dimensions in sibling groups being invalid per NetCDF-4 helps with one of the two issues here. I also believe the updated implementation in the PR addresses the other case. I'll resolve this comment. Thanks, both.

owenlittlejohns (changes requested):
Thanks for making further updates. I think there are two things I spotted in the updated code:

1) `all_dimension_sizes` is only declared on `VarInfoFromDmr`, but you are populating it in `VarInfoFromNetCDF4`.
2) The comparison of the number of dimensions with a shape from a `<Dim />` element needs to be against the total number of `<Dim />` elements for that variable only, not all entries in the `all_dimension_sizes` cache.

Both should be quick fixes and I'm happy to chat through them on Slack if that will help.

varinfo/var_info.py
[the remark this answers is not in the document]
owenlittlejohns: The update to use a class attribute results this issue; resolving the comment now. Thanks!

tests/unit/test_variable.py
owenlittlejohns: I think this has also been addressed. Thanks!

varinfo/var_info.py
owenlittlejohns: Going to mark this as resolved. I was getting mixed up by the `VarInfoBase` having the class attribute, but the attribute only being passed to `VariableFromDmr`. Sorry about that!

varinfo/variable.py
owenlittlejohns: Nice. Marking conversation thread as resolved.

Issue #56, opened 2025-04-23
Support downloading a DMR file in earthdata-varinfo


eni-awowale:
Currently, `earthdata-varinfo` does not support reading HDF4 files.

To work around this, users can download a DMR file and open it with `VarInfoFromDmr`, and then call `get_all_umm_var` on the `VarInforFromDmr` object to generate UMM-Var records.

It would nice if `earthdata-varinfo` supported directly downloading DMR files like it does for downloading granule data download URLs (see "RelatedUrls" and "Get Data" attributes in a granule metadata record).

To download a DMR file, `earthdata-varinfo` can go into a granule metadata record, utilize the OPeNDAP data url and append `.dmr.xml` to get the DMR response as an xml. And then use a standard library function like `xml.etree.ElementTree` to write out the response.

Example:
`https://acdisc.gesdisc.eosdis.nasa.gov/opendap/Aqua_AIRS_Level3/AIRS3STD.006/2002/AIRS.2002.08.31.L3.RetStd_IR001.v6.0.9.0.G13214161357.hdf.dmr.xml`

D-Auty:
I assume you are referring to the existing generate_collection_umm_var method, that has some heuristic for downloading NetCDF files and producing UMM_Var records.  I assume you are referring to a keyword argument something like “use_dmr= False” (False being the current default behavior). Setting to True would imply downloading DMR files, rather than downloading the whole granule.  That makes sense to me.

owenlittlejohns:
Having a keyword argument to grab a DMR file sounds like a good option to me. We'd basically be able to pass around a `VarInfoFromDmr` instance in place of a `VarInfoFromNetCDF4`. That feels like a pretty quick change (famous last words).

I feel like we'd probably want a default value, and that should probably be to use the data file instead of a DMR. Does anyone have any strong opinions on the default behaviour?

PR #60, opened 2025-06-27
Support downloading a DMR file in earthdata-varinfo 

eni-awowale:
## Description
See #56. Currently leaving as draft PR for hackfest

## PR Acceptance Checklist
* [ n/a ] Jira ticket acceptance criteria met.
* [ n/a] `VERSION` updated if publishing a release.


eni-awowale:
@owenlittlejohns and @flamingbear could you all drop a review when you are free? I still need to update the docs though.

varinfo/generate_umm_var.py
```diff
@@ -64,20 +66,35 @@ def generate_collection_umm_var(
...
 
-        # Parse the granule with VarInfo to map all variables and relations:
-        var_info = VarInfoFromNetCDF4(local_granule, config_file=config_file)
+            # Parse the granule with VarInfo to map all variables and relations:
+            var_info = VarInfoFromNetCDF4(local_granule, config_file=config_file)
 
-        # Generate all the UMM-Var records:
-        all_umm_var_records = get_all_umm_var(var_info)
+            # Generate all the UMM-Var records:
+            all_umm_var_records = get_all_umm_var(var_info)
```
owenlittlejohns, line 97: It feels like thre's too much in the scope of this `if`/`else`. There are really only two things that are different for each branch:

1) The function to get the `granule_link`.
2) The class to generate `var_info`.

It might be cleaner to only put what is different in you conditions. Something like:

```
if use_dmr:
    # Get OPeNDAP DMR by appending `.dml.xml` to the OPeNDAP URL
    granule_link = get_dmr_xml_url(granule_response)
else:
    # Get the data download URL for the most recent granule (NetCDF-4 file)
    granule_link = get_granule_link(granule_response)

with TemporaryDirectory() as temp_dir:
    # Download file to runtime environment
    local_granule = download_granule(
        granule_link, auth_header_edl_token, out_directory=temp_dir
    )

    if use_dmr:
        var_info = VarInfoFromDmr(local_granule, config_file=config_file)
    else:
        # Parse the granule with VarInfo to map all variables and relations:
        var_info = VarInfoFromNetCDF4(local_granule, config_file=config_file)

# Generate all the UMM-Var records:
all_umm_var_records = get_all_umm_var(var_info)
```

varinfo/cmr_search.py
```diff
@@ -190,3 +190,21 @@ def get_edl_token_header(auth_header: str, cmr_env: CmrEnvType) -> str:
...
+
+
+def get_dmr_xml_url(granule_response: Sequence) -> str:
+    """Gets the OPeNDAP url from a CMR granule response and appends `.dmr.xml`"""
+    granule_link = next(
+        (
+            link['href']
+            for link in granule_response[0].get('links', [])
+            if link['rel'].endswith('/service#')
+            and 'inherited' not in link  # and 'opendap' in link
```
owenlittlejohns, line 202: Should that last condition be commented out? (It feels like it should either be enabled or removed)

If it is removed, how do we know that it's only an OPeNDAP service link that is being retrieved, what about other service links? What are the things that make an OPeNDAP URL unique in the `.json` response from CMR? (I could tell you for the `.umm_json` response, but that's not what we're dealing with here)

tests/unit/test_generate_umm_var.py
```diff
@@ -20,6 +20,13 @@ def setUpClass(cls):
         cls.launchpad_token_header = 'launchpad-token'
         cls.netcdf4_basename = 'f16_ssmis_20210426v7.nc'
         cls.netcdf4_url = f'https://example.com/{cls.netcdf4_basename}'
+        cls.example_dmr_basename = 'M2I3NPASM_example.dmr'
+        cls.opendap_url = (
+            f'https://fake.opendap.earthdata.nasa.gov/{cls.example_dmr_basename}'
+        )
+        cls.opendap_xml_download_url = (
+            f'https://fake.opendap.earthdata.nasa.gov/{cls.example_dmr_basename}.xml'
```
owenlittlejohns, line 28: Should this be `.xml` or `.dmr.xml`?

owenlittlejohns (changes requested):
Generally these changes look good. I had mainly two comments:

1) (More important) - How to be sure it's only OPeNDAP service URLs that are retrieved (we'll likely need a unit test that shows that, too).
2) Some nitpicking with making sure that the `if`/`else` in the main retrieval doesn't just encapsulate everything when it doesn't need to.

owenlittlejohns (changes requested):
I think this looks good. I'm really only requesting one change, and it's an easy one: I think you should publish a new version of the package with this functionality. You'd just want to:

* Update `VERSION` to have a new version number. I'd recommend making this 3.2.0 as you're adding a new optional argument to a public-facing function.
* Tweak the `CHANGELOG.md` to say v3.2.0 instead of v3.1.1. Also add the date in place of "Unreleased".

Then I'll hit approve, you can merge things, and this will hit PyPI via the CI/CD. Nice work!

tests/unit/test_cmr_search.py
```diff
@@ -519,3 +520,120 @@ def test_get_edl_token_header_with_edl_token(self):
...
+                        'title': 'OPeNDAP request URL (GET DATA : OPENDAP DATA)',
+                        'hreflang': 'en-US',
+                        'href': 'https://fake.earthdata.nasa.gov/example.nc4',
+                    }
+                ]
+            }
+        ]
+
+        with self.subTest('Granule has no `links` key'):
+            with self.assertRaises(MissingGranuleDownloadLinks) as context_manager:
```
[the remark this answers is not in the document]
eni-awowale: Nice thanks! This is good to know for next time. I am more familiar with using the built in pattern matching with pytest.

PR #102, opened 2026-05-04
TRT-800: Add UserWarning when generating UMM-Var records from a DMR.

owenlittlejohns:
## Description

This PR updates `generate_collection_umm_var` to emit a `UserWarning` when generating UMM-Var records from a DMR file.

_Open question_ - I chose a `UserWarning`. Another option is a `RuntimeWarning`. Here are the definitions (from the [documentation](https://docs.python.org/3/library/warnings.html#warning-categories)):

* `UserWarning`: The default category for `warn()`.
* `RuntimeWarning`: Base category for warnings about dubious runtime features.

I'm open to either option. Any thoughts?

## Jira Issue ID

TRT-800

## Local Test Steps

Using this branch, and a conda environment with all necessary dependencies, run the following code locally:

```
from varinfo.generate_umm_var import generate_collection_umm_var

umm_var_json = generate_collection_umm_var(collection_concept_id='C1245618475-EEDTEST', auth_header='...', publish=False, use_dmr=True)
```

The `auth_header` could be either a valid Launchpad token or an EDL bearer token (with the `Bearer` prefix).

You should see the following output to the terminal:

```
/path/on/your/local/machine/to/earthdata-varinfo/varinfo/generate_umm_var.py:76: UserWarning: Generating UMM-Var records from a DMR document will replace spaces in variable paths with underscores. Please use the source data file to generate UMM-Var instead, where possible.
  warn(DMR_GENERATION_WARNING, UserWarning)
```

To check that the warning is not issued when not using a DMR to generate the UMM-Var records:

```
from varinfo.generate_umm_var import generate_collection_umm_var

umm_var_json = generate_collection_umm_var(collection_concept_id='C1245618475-EEDTEST', auth_header='...', publish=False, use_dmr=False)
```

You should not see the warning in your terminal this time.

## PR Acceptance Checklist


flamingbear (approved):
Does what it says. 

```
>>> umm_var_json = generate_collection_umm_var(collection_concept_id='C1245618475-EEDTEST', auth_header=auth, publish=False, use_dmr=True)
/Users/savoie/projects/data-services/earthdata-varinfo/varinfo/generate_umm_var.py:76: UserWarning: Generating UMM-Var records from a DMR document will replace spaces in variable paths with underscores.
  warn(DMR_GENERATION_WARNING, UserWarning)
>>> umm_var_json = generate_collection_umm_var(collection_concept_id='C1245618475-EEDTEST', auth_header=auth, publish=False, use_dmr=False)
>>> 
```

D-Auty:
I might add the words " - against policy" to the warning. I'm not hard over on this, and the wording is a compromise between saying nothing and providing more explanation. My concern is that as stated, a curator might think replacing blanks with "_" is actually a good idea. Perhaps instead of those words, a reference to a policy statement that provides more explanation?

owenlittlejohns:
@D-Auty a couple of quick thoughts here:

We probably can't use the word "policy" - that was a term identified in January 2025 as referring to something only civil servants can make. (A pedantic observation, I know, but it was one of the terms we were asked to remove from repositories)

Generally warnings are emitted when moderately undesired or questionable (but not explicitly prohibited) behaviour is occurring. So it's hopefully unlikely a user would/should think that the presence of a warning is instructing them of good behaviour in code.

That said, you're definitely right that the warning is a bit of a compromise, and that it doesn't provide clear instructions for what to do instead, which would definitely be helpful. How about adding the following:

> Please use the source data file to generate UMM-Var instead, where possible.

This would make the whole warning:

> UserWarning: Generating UMM-Var records from a DMR document will replace spaces in variable paths with underscores. Please use the source data file to generate UMM-Var instead, where possible.

owenlittlejohns:
I've updated the warning message in [f1a68fa](https://github.com/nasa/earthdata-varinfo/pull/102/changes/f1a68fa8eaac5485aeb1b633f579936d328fb61f). It is now as suggested above:

> UserWarning: Generating UMM-Var records from a DMR document will replace spaces in variable paths with underscores. Please use the source data file to generate UMM-Var instead, where possible.

owenlittlejohns:
Following the message update, I was going to merge this tomorrow (7th May), unless there are objections.

PR #103, opened 2026-05-14
CVE-2026-44432: Upgrade python-cmr to prevent vulnerability.

flamingbear:
## Description
* Sets `python-cmr >=0.13.0` to end direct `urllib3` override and prevent CVE-2026-44432.

## Jira Issue ID

vulnerability 

## Local Test Steps

``` shell
❯ make build
❯ make test
```

## PR Acceptance Checklist
* [ ] ~Jira ticket acceptance criteria met.~
* [X] `CHANGELOG.md` updated to include high level summary of PR changes.
* [X] `VERSION` updated if publishing a release.
* [X] Tests added/updated and passing.
* [ ] ~Documentation updated (if needed).~


requirements.txt
```diff
@@ -1,6 +1,6 @@
 netCDF4 >= 1.7.2
 numpy >= 1.24.2,<2.3
-python-cmr ~= 0.12.0
-requests ~= 2.31.0
-# To handle CVE-2025-66471 (coming from python-cmr)
-urllib3 ~= 2.6.1
+python-cmr >= 0.12.0
+requests >= 2.31.0
```
owenlittlejohns: Snyk also reckons there are issues with `requests`. There are three different medium severity CVEs; making this `requests >= 2.33.0` would squash those, too, I'd guess.

(I tested under `urllib3 ~= 2.7.0` and `requests ~= 2.33.0` and things ran smoothly locally, so I'd guess using `>=` should also work.)

requirements.txt
```diff
@@ -1,6 +1,6 @@
...
+# To handle CVE-2025-66471 (coming from python-cmr) and CVE-2026-44432
+urllib3 >= 2.7.0
```
owenlittlejohns: I wonder if we can just remove this by bumping `python-cmr` to `>=0.13.0`? I tried just installing `python-cmr` with its latest version, and that was `python-cmr==0.13.0` and pulled in `urllib3==2.7.0` by default.

It would be nice to cut out this explicit dependency if we could.

owenlittlejohns (changes requested):
Looks great (very similar to the PR I was about to open 😉). There are a couple of small changes suggested:

* A bump for requests, too.
* Possibly bumping `python-cmr`, which might mean we can stop specifying `urllib3` at all.

Then I think we can :shipit: 

I guess there should probably be a flurry of PRs following this to bump `earthdata-varinfo` in all our services. 😬

owenlittlejohns:
Oh shoot. I just saw the Python 3.9 error (which you even mentioned in Slack). We should talk about whether we need to support Python 3.9. It was EOL 31st October 2025, so I could easily be convinced to cut that from the CI/CD (with a quick update to `setup.py` maybe to indicate what is supported)

requirements.txt
flamingbear: That's a great thought.  I'll do that. 7b6dbc2

owenlittlejohns:
I updated the CI/CD in #104. If we have `urllib3 >= 2.7.0` and lose 3.9 compatibility, should this release be more than a patch version? (No massively strong opinions, but it is a breaking change - albeit for an EOL Python version)

owenlittlejohns (approved):
This looks good to me 🚀 

Interesting that the 3.14 build is a bit slower than the rest, but that's not down to us. (Something something `uv` in the future?)

PR #105, opened 2026-05-15
DAS-2432 Adds functions to identify and retrieve excluded science variables.

vutrannasa:
## Description

This is to just abstract some repeated code across different repositories that checks whether of not a variable is in the service's `ExcludeScienceVariables` `earthdata-varinfo` configuration file. 
 
And will replace the duplicate functions in the [harmony-regridding-service](https://github.com/nasa/harmony-regridding-service/blob/47272929ae175959c1d8f88c8ca06edef42dbaed/harmony_regridding_service/var_utilitities.py#L31C1-L36C66), [harmony-metadata-annotator](https://github.com/nasa/harmony-metadata-annotator/blob/257cbc9183f5f6f4c8509b256ba6f630083fe6ce/metadata_annotator/annotate.py#L535). and [hoss](https://github.com/nasa/harmony-opendap-subsetter/blob/418e2228fd4771ebbeb5899497a0e9b2e41f4baa/hoss/variable_utilities.py#L70) (update link when [DAS-2417](https://bugs.earthdata.nasa.gov/browse/DAS-2417) is merged), and remove the associated unit tests in each repository.

### Fix
1) Add `get_excluded_science_variables()`
    Return the set of excluded science variables in VarInfo configuration .json
2) Add `is_excluded_science_variable()`
    Return True/False if science variable is exclude

## Jira Issue ID
https://bugs.earthdata.nasa.gov/browse/DAS-2432

## Local Test Steps
```
$ git clone https://github.com/nasa/earthdata-varinfo.git
$ cd earthdata-varinfo
$ git DAS-2432-is_excluded_science_variable
$ make test
```

## PR Acceptance Checklist
* [ x] Jira ticket acceptance criteria met.
* [ x] `CHANGELOG.md` updated to include high level summary of PR changes.
* [ x] `VERSION` updated if publishing a release.
* [ x] Tests added/updated and passing.
* [ x] Documentation updated (if needed).


varinfo/var_info.py
```diff
@@ -150,6 +151,12 @@ def _set_short_name(self):
 
         """
 
+    def _set_exclusions_pattern(self) -> re.Pattern:
```
vutrannasa, line 153: Compile and store the exclusion pattern once to avoid duplicate `re.compile` calls

Duplicate code from 
[harmony-regridding-service](https://github.com/nasa/harmony-regridding-service/blob/47272929ae175959c1d8f88c8ca06edef42dbaed/harmony_regridding_service/var_utilitities.py#L31C1-L36C66)
[harmony-metadata-annotator](https://github.com/nasa/harmony-metadata-annotator/blob/257cbc9183f5f6f4c8509b256ba6f630083fe6ce/metadata_annotator/annotate.py#L535)
[hoss](https://github.com/nasa/harmony-opendap-subsetter/blob/418e2228fd4771ebbeb5899497a0e9b2e41f4baa/hoss/variable_utilities.py#L70)

varinfo/var_info.py
```diff
@@ -168,15 +175,11 @@ def get_variables_with_coordinates(self) -> dict[str, OutputVariableType]:
         variable in the configuration file supplied to the object.
 
         """
-        exclusion_pattern = re.compile(
```
vutrannasa, line 171: Remove duplicate code

varinfo/var_info.py
```diff
@@ -214,26 +217,46 @@ def is_science_variable(self, variable: OutputVariableType) -> bool:
 
         return False
 
+    def is_excluded_science_variable(self, variable_path: str) -> bool:
```
vutrannasa, line 217: Return True/False if science variable is exclude

varinfo/var_info.py
```diff
@@ -214,26 +217,46 @@ def is_science_variable(self, variable: OutputVariableType) -> bool:
...
             for variable_path, variable in self.variables.items()
             if variable_path is not None
             and self.is_science_variable(variable)
-            and not self.variable_is_excluded(variable_path, exclusions_pattern)
+            and not self.variable_is_excluded(variable_path, self.exclusions_pattern)
         }
 
         return filtered_with_coordinates - self.references
 
+    def get_excluded_science_variables(self) -> set[str]:
```
vutrannasa, line 242: Return the set of excluded science variables

vutrannasa (commented):
[HOSS development branch ](https://github.com/nasa/harmony-opendap-subsetter/commit/b75ad3de8fa86ea1ae4954f2395920e3caea448b)to use earthdata-varinfo 5.1.0 `get_excluded_science_variables()`

[Metadata Annotator development branch ](https://github.com/nasa/harmony-metadata-annotator/commit/74832f812f9ae5580598769190f79d14fa5000b2)to use earthdata-varinfo 5.1.0 `get_excluded_science_variables()`

[harmony-regridding-service development branch ](https://github.com/nasa/harmony-regridding-service/compare/main...DAS-2432-is_excluded_science_variable)

D-Auty (approved):
Updates look fine

make test ran successfully, no failures, 99% stmt coverage.

README.md
```diff
@@ -53,6 +53,9 @@ var_info.get_science_variables()
 # Retrieve a set of variables without coordinate metadata:
 var_info.get_metadata_variables()
 
+# Retrieve a set of excluded variables defined in JSON configuration file:
+var_info.get_excluded_science_variables()
```
owenlittlejohns, line 57: Thanks for updating the README with this information!

varinfo/var_info.py
owenlittlejohns: Nit: This type hint is incorrect - the method is not actually returning anything, it's assigning an attribute of the class within the method.

I think there are two options here:

1) Just fix the type hint to have a `None` return type.
2) Actually make the method return the value, and assign it to `self.exclusions_pattern` directly in the `__init__`, instead of having `self.exclusions_pattern = None`.

I started thinking about the way this is happening for other class attributes in the `__init__` method when I was hoping to include `mypy` checks for `earthdata-varinfo`. I think those statements in the `__init__` were one of the places that `mypy` complained about (a lot 😬), and it's probably right.

The choice between option 1 and 2 above probably boils down to: do we want to be consistent with what's already there, or do we want to start moving to potentially better code. I'd be fine with whichever, but have a mild preference to option 2.

varinfo/var_info.py
```diff
@@ -150,6 +151,12 @@ def _set_short_name(self):
 
         """
 
+    def _set_exclusions_pattern(self) -> re.Pattern:
+        """Compile the exclusion regex pattern."""
+        self.exclusions_pattern = re.compile(
+            '|'.join(self.cf_config.excluded_science_variables)
+        )
```
owenlittlejohns: This is a more efficient way to do things, compiling the regular expression pattern only the once. Thanks for the clean-up.

tests/unit/test_var_info.py
```diff
@@ -488,6 +499,21 @@ def test_var_info_variable_is_excluded(self):
...
+
+        """
+        dataset = VarInfoFromDmr(self.mock_dmr_two, config_file=self.test_config_file)
+        self.assertTrue(
+            dataset.is_excluded_science_variable('/exclude_one/has_coordinates')
+        )
+        self.assertFalse(
+            dataset.is_excluded_science_variable('/required_group/has_no_coordinates')
+        )
+        self.assertFalse(dataset.is_excluded_science_variable('/science/lat_bnds'))
```
owenlittlejohns: Nit - these could each be in their own `self.subTest` context managers. It gives a little bit of better isolation for each test case. If your first test case fails, I think `self.subTest` ensures the other sub tests will still run. Right now, the first failed assertion will stop the other assertions from being checked.

This would give the opportunity to add a short description for each test case, so it's clear why each variable is being tested.

tests/unit/test_var_info.py
```diff
@@ -419,6 +419,17 @@ def test_var_info_get_science_variables(self):
...
+    def test_var_info_get_excluded_science_variables(self):
+        """Ensure the correct set of excluded science variables is returned. This
+        should account for excluded science variables defined in the
+        associated instance of the `CFConfig` class.
+
+        """
+        dataset = VarInfoFromDmr(self.mock_dmr_two, config_file=self.test_config_file)
+
+        excluded_science_variables = dataset.get_excluded_science_variables()
+        self.assertEqual(excluded_science_variables, {'/exclude_one/has_coordinates'})
```
owenlittlejohns: You could/should use [assertSetEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSetEqual), like you do elsewhere in the PR.

Secondary, super nit - it's redundant to declare a variable and then only use it once on the next line.

Summed up, you could probably just do:

```
self.assertSetEqual(
    dataset.get_excluded_science_variables(),
    {'/exclude_one/has_coordinates'},
)
```

owenlittlejohns (changes requested):
Nice work @vutrannasa - this is some great clean-up!

I love that this pulls duplicate code from across multiple services into one place, and that one place really makes sense to be here. Nice! Also, the clean-up to compile the regular expression pattern once is a nice optimisation.

I only had nits. One is a bit of a can of worms, and entirely down to the way I declared class attributes way-back-when. Anything other than `self.exclusions_pattern` is _way_ out of scope, though. (If you want to just tweak the type hint and avoid any of this for now, I'd be totally fine with that, so long as the type hint is matching what's actually happening)

tests/unit/test_var_info.py
vutrannasa: Update to [assertSetEqual](https://github.com/nasa/earthdata-varinfo/pull/105/changes/8c75f7031a03a0dfd933c3ec32ce2a5ea950dbbf#diff-844e20787aaad533c9fe687b0ba40320d60c7f5a30e54a3a3116067b5882ca2bR430)

tests/unit/test_var_info.py
vutrannasa: Update to use [self.subTest ](https://github.com/nasa/earthdata-varinfo/pull/105/changes/8c75f7031a03a0dfd933c3ec32ce2a5ea950dbbf#diff-844e20787aaad533c9fe687b0ba40320d60c7f5a30e54a3a3116067b5882ca2bR511)with short description

varinfo/var_info.py
vutrannasa: In `__init__` assign to [self.exclusions_pattern = self._set_exclusions_pattern()](https://github.com/nasa/earthdata-varinfo/pull/105/changes/8c75f7031a03a0dfd933c3ec32ce2a5ea950dbbf#diff-750ca18cc96858ef32be7e9796bde2829b8da5fbbd7dd02aa94d87d8cad9bf4dR75)

owenlittlejohns: Thanks, @vutrannasa - I hope this is the start of tidying up the `__init__` generally!

owenlittlejohns (approved):
Thanks for the quick update - let me know if you need eyes on the follow-up PRs to use this version of `earthdata-varinfo` in services. Happy to give you some reviews to wrap up this really nice clean-up!

